PR #67, opened 2021-11-17
Fix flakey test (issue 45)

jingibus:
I was unable to get the original test to repro its flakiness, but on the principle that `TestCoroutineDispatcher` is not to be trusted unless the circumstances absolutely require it, I rewrote the test without `TestCoroutineDispatcher` and got a consistent failure. Fixed this failure by moving up the Composition hookup inside `launchMolecule`.


jingibus:
Superseded by #71.

PR #70, opened 2021-11-21
New Sample : Fully Compose Activity

theapache64:
## Changes

- Adds a new fully compose activity sample (without XML/viewBinding)

## Why do we need this PR?

The main target of this PR is not to get it merged, but more like "review-only" or for "later-reference". But if you think this will be useful, I'd be much happier to see this merged 😉 

## Output

https://user-images.githubusercontent.com/9678279/142777319-3d937a5e-8428-4a22-b6d6-000e42841d07.mp4

### DIFF
--- a/sample/build.gradle
+++ b/sample/build.gradle
@@ -7,6 +7,8 @@ dependencies {
   implementation 'com.squareup.retrofit2:converter-scalars:2.9.0'
   implementation 'com.squareup.okhttp3:okhttp:4.9.2'
   implementation 'com.squareup.okhttp3:logging-interceptor:4.9.2'
+  implementation "androidx.activity:activity-compose:1.4.0"
+  implementation "androidx.compose.material:material:1.0.4"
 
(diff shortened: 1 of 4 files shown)


JakeWharton:
There's no reason to use Moleule with Compose UI like this. You can simply call the function from your Compose UI tree or from a function which binds your presenter to your UI as a composable. You may still use Molecule's testing artifact for testing that function, but you wouldn't use Molecule in the app.

The only reason you would use Molecule and Compose UI was if you had an architectural split between your presenter and rendering layers which was enforced through some kind of interface.  For example, our presenters all conform to roughly this interface:
```kotlin
interface Presenter<E, M> {
  fun present(events: Flow<E>): StateFlow<M>
}
```
This allows us to use RxJava, Flow, or Molecule as an implementation detail of a presenter. Additionally, from the render layer we're free to use either Compose UI or Android Views with any presenter.

theapache64:
Ahaa.. makes sense. Thanks for the review :) Closing this PR

PR #71, opened 2021-12-03
Fix bug in frame triggering for BroadcastFrameClock contexts

### LINKED ISSUE #45: Flaky test: MoleculeTest.errorDelayed
```
 app.cash.molecule.MoleculeTest > errorDelayed FAILED
    java.lang.AssertionError: expected:<1> but was:<0>
        at org.junit.Assert.fail(Assert.java:89)
        at org.junit.Assert.failNotEquals(Assert.java:835)
        at org.junit.Assert.assertEquals(Assert.java:120)
        at org.junit.Assert.assertEquals(Assert.java:146)
        at app.cash.molecule.MoleculeTest.errorDelayed(MoleculeTest.kt:126)
```

jingibus:
`Recomposer.runRecomposeAndApplyChanges` always waits for a frame tick before recomposing a child composition. As a result, the call to `Snapshot.sendApplyNotifications()` isn't achieving the desired outcome in the existing code: under test, writes to snapshot state will fail to produce an additional output. `snapshotUpdatesTickEnclosingClock` illustrates the issue.

To get the test to pass, though, the call to `composition.setContent` has to be moved to the beginning of `launchMolecule`. This is probably due to a race within `Recomposer`.

Also fixes #45.

### DIFF
--- a/molecule/molecule-runtime/src/commonTest/java/app/cash/molecule/MoleculeTest.kt
+++ b/molecule/molecule-runtime/src/commonTest/java/app/cash/molecule/MoleculeTest.kt
@@ -22,19 +22,26 @@ import androidx.compose.runtime.mutableStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
 import androidx.compose.runtime.snapshots.Snapshot
+import kotlin.coroutines.CoroutineContext
+import kotlin.test.assertFailsWith
+import kotlin.time.ExperimentalTime
 import kotlinx.coroutines.CoroutineExceptionHandler
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.ExperimentalCoroutinesApi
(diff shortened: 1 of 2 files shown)


DSteve595:
Sounds like the same behavior I was seeing [here](https://github.com/cashapp/molecule/discussions/64#discussioncomment-1647720).

jingibus:
Maybe. These changes don't break `CounterPresenterTest`; however, now I'm realizing that `testMolecule` (which we're using in the code I modeled this repro test off of) is also sending a frame. So this may not actually fix my issue. 🤔 🤔 🤔 🤔 

In any event, I think that the expectation in the test case is reasonable; it would be really confusing in prod if that scenario failed to produce an additional model.

jingibus:
Pulling until I work out the issues with this PR. It doesn't address my issue as I thought it did, so: needs more work

PR #74, opened 2022-02-07
Dispose composition after job is completed

mattprecious:
This enables a DisposableEffect's onDispose to be invoked.

### DIFF
--- a/molecule/molecule-runtime/src/commonTest/java/app/cash/molecule/MoleculeTest.kt
+++ b/molecule/molecule-runtime/src/commonTest/java/app/cash/molecule/MoleculeTest.kt
@@ -16,6 +16,7 @@
 package app.cash.molecule
 
 import androidx.compose.runtime.BroadcastFrameClock
+import androidx.compose.runtime.DisposableEffect
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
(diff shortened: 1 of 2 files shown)


molecule/molecule-runtime/src/main/java/app/cash/molecule/molecule.kt
```diff
@@ -105,6 +105,7 @@ fun <T> CoroutineScope.launchMolecule(
     }
   }
   coroutineContext.job.invokeOnCompletion {
+    composition.dispose()
```
mattprecious, line 108: Not sure if there's any significance to the ordering here...

JakeWharton: It's all single-threaded so order doesn't matter.

molecule/molecule-runtime/src/commonTest/java/app/cash/molecule/MoleculeTest.kt
```diff
@@ -202,4 +203,29 @@ class MoleculeTest {
...
   }
+
+  enum class DisposableEffectState { NOT_LAUNCHED, LAUNCHED, DISPOSED }
+
+  @Test fun disposableEffectDisposesWhenScopeIsCancelled() {
+    val dispatcher = TestCoroutineDispatcher()
+    val clock = BroadcastFrameClock()
+    val scope = CoroutineScope(dispatcher + clock)
+
+    var state: DisposableEffectState = DisposableEffectState.NOT_LAUNCHED
```
JakeWharton, line 214: I usually use an `Int` which starts at 0 and then is incremented but this is fine (and not in the meme way).

JakeWharton (approved):

PR #77, opened 2022-02-16
Add expectNoEvents to MoleculeTurbine

abyrnes:
Please close if this is silly or useless, but I don't see another way of verifying this behavior other than expecting a timeout, which extends the runtime of a test by a full second.

### DIFF
--- a/molecule/molecule-testing/src/commonTest/java/app/cash/molecule/testing/MoleculeTestingTest.kt
+++ b/molecule/molecule-testing/src/commonTest/java/app/cash/molecule/testing/MoleculeTestingTest.kt
@@ -37,7 +37,6 @@ import org.junit.Assert.assertTrue
 import org.junit.Ignore
 import org.junit.Test
 import kotlin.test.assertFailsWith
-import kotlin.test.fail
 import kotlin.time.Duration
 import kotlin.time.ExperimentalTime
 
@@ -271,12 +270,8 @@ class MoleculeTestingTest {
   @Test
   fun expectNoEventsFail() {
     testMolecule({}) {
-      try {
-        expectNoEvents()
-        fail()
-      } catch (t: AssertionError) {
-        assertEquals(t.message, "Expected no events but found Item(kotlin.Unit)")
-      }
+      val t = assertFailsWith<AssertionError> { expectNoEvents() }
+      assertEquals(t.message, "Expected no events but found Item(kotlin.Unit)")
     }
   }
 }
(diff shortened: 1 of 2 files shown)


molecule/molecule-testing/src/commonTest/java/app/cash/molecule/testing/MoleculeTestingTest.kt
```diff
@@ -258,4 +259,24 @@ class MoleculeTestingTest {
...
+
+  @Test
+  fun expectNoEventsFail() {
+    testMolecule({}) {
+      try {
+        expectNoEvents()
+        fail()
+      } catch (t: AssertionError) {
+        assertEquals(t.message, "Expected no events but found Item(kotlin.Unit)")
+      }
```
JakeWharton: Please use `assertFailsWith` rather than try/fail/catch. Right now you're getting lucky in that the `AssertionError` that would be thrown by `fail()` is being caught but is failing in the message check. Would rather using something a bit safer.

JakeWharton (approved):

JakeWharton:
This exists in Turbine so I'm okay duplicating it here.

I'll just comment once again for the record that this assertion is prone to race conditions and does not guarantee that no events will _ever_ be produced in the future (anywhere from 1ms to the heat death of the universe). It has it's place, but no doubt someone will get upset when it's doesn't do exactly what they think.

molecule/molecule-testing/src/commonTest/java/app/cash/molecule/testing/MoleculeTestingTest.kt
abyrnes: Done!

abyrnes:
> This exists in Turbine so I'm okay duplicating it here.
> 
> I'll just comment once again for the record that this assertion is prone to race conditions and does not guarantee that no events will _ever_ be produced in the future (anywhere from 1ms to the heat death of the universe). It has it's place, but no doubt someone will get upset when it's doesn't do exactly what they think.

Would it be worthwhile to mention this in the kdoc?

mattprecious:
I don't think this is ever a valid assertion to make given the current implementation of the testing library. Since recomposition is delayed until you call `awaitItem()`, there's no way for `expectNoEvents()` to fail in any case except the first synchronous composition, which will always emit an item.

We've moved off of this testing library internally and are experimenting with a more aggressive recomposition strategy. Once that implementation settles I think we can push it upstream here.

JakeWharton:
Ah, yes. I thought we had a reason for not including it I just couldn't... remember.

PR #80, opened 2022-03-04
Migrate API to use Duration

JakeWharton:
As of Kotlin 1.6 this type is now stable.

### DIFF
--- a/molecule/molecule-testing/src/main/java/app/cash/molecule/testing/moleculeTesting.kt
+++ b/molecule/molecule-testing/src/main/java/app/cash/molecule/testing/moleculeTesting.kt
@@ -33,20 +33,19 @@ import kotlinx.coroutines.withTimeout
 import kotlinx.coroutines.yield
 import kotlin.coroutines.CoroutineContext
 import kotlin.time.Duration
-import kotlin.time.ExperimentalTime
-
-@ExperimentalCoroutinesApi
-@ExperimentalTime
-fun <T> testMolecule(
-  body: @Composable () -> T,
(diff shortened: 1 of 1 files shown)


molecule/molecule-testing/src/main/java/app/cash/molecule/testing/moleculeTesting.kt
```diff
@@ -33,20 +33,19 @@ import kotlinx.coroutines.withTimeout
 import kotlinx.coroutines.yield
 import kotlin.coroutines.CoroutineContext
 import kotlin.time.Duration
-import kotlin.time.ExperimentalTime
+import kotlin.time.Duration.Companion.milliseconds
 
 @ExperimentalCoroutinesApi
-@ExperimentalTime
 fun <T> testMolecule(
```
JakeWharton, line 39: Should this `Long`-accepting overload be deprecated? I don't have any good arguments to keep it long-term.

oldergod (approved):

molecule/molecule-testing/src/main/java/app/cash/molecule/testing/moleculeTesting.kt
JakeWharton: A thumb is not an argument

mattprecious: We should probably just replace this whole module with what we're using now internally...

mattprecious (approved):

PR #90, opened 2022-05-14
Add RecompositionClock param

jingibus:
This PR is an upstream of `backpressureMoleculeFlow`, a tool we have been using internally at Cash App for Rx interop, but mostly for testing. We avoid molecule's testing tool suite entirely and use `backpressureMoleculeFlow` with Turbine (along with a `distinctUntilChanged()` call).

I can't think of any good reason not to have a call to `distinctUntilChanged()` when consuming a Molecule, but I've left it out here anyway.


molecule/molecule-runtime/src/commonMain/kotlin/app/cash/molecule/GatedFrameClock.kt
```diff
@@ -0,0 +1,42 @@
...
+    set(value) {
+      val started = value && !field
+      field = value
+      if (started) {
+        sendFrame()
+      }
+    }
+
+  private fun sendFrame() {
+    clock.sendFrame(start.elapsedNow().inWholeNanoseconds)
```
mattprecious, line 43: Is this time needed for anything? The existing testing code just sends `0` because it doesn't matter.

mattprecious (approved):
Needs a `spotlessApply`.

molecule/molecule-runtime/src/commonMain/kotlin/app/cash/molecule/GatedFrameClock.kt
jingibus: Not that I know of. Is there any downside to providing a meaningful value?

mattprecious: The "downside" is that this is doing work that it might not need to do 🤷 I don't know enough about the internals to say whether this is truly meaningless, meaningless outside of UI, or meaningful in some subtle way.

jingibus: In JVM land, this would no doubt turn into a `System.nanoTime()` in the `actual` implementation. So I can't imagine it's too expensive.

JakeWharton:
Out all week. Will look next!

JakeWharton:
Okay so upon discussing this a bit with @swankjesse and @mattprecious here's our take:

`moleculeFlow` is a cold `Flow` factory and its API naming is in the form of `callbackFlow` or `channelFlow` with the suffix.

`launchMolecule` is a hot `StateFlow` factory and its API naming is in the form of `launchIn` or just a plain hot `launch` with the prefix.

This PR is a new axis which is whether or not the frame clock is extracted from the user context or eagerly ticks automatically.

> FrameClock => rate-limited or eager
> Launch mode => "hot" (launch) or "cold" (Flow)

So the proposal is to introduce a
```kotlin
enum RecompositionMode {
  FrameLimited, // batched, user-controlled
  Immediate, // eagerly recomposed
}
```
which dictates the behavior of the frame clock as a parameter on both functions. This parameter will have no default so the caller is forced to learn what each means.

Thoughts?

jingibus:
> This PR is a new axis which is whether or not the frame clock is extracted from the user context or eagerly ticks automatically.
> 
> > FrameClock => rate-limited or eager
> > Launch mode => "hot" (launch) or "cold" (Flow)
> 
> So the proposal is to introduce a
> 
> ```kotlin
> enum RecompositionMode {
>   FrameLimited, // batched, user-controlled
>   Immediate, // eagerly recomposed
> }
> ```
> which dictates the behavior of the frame clock as a parameter on both functions. This parameter will have no default so the caller is forced to learn what each means.
> 
> Thoughts?

The direction makes sense to me. `FrameLimited, // batched, user-controlled` is confusing, though: from our discussion here, I know that this means "use whatever clock happens to already be in the context (which we expect to be the Android frame clock)", but `FrameLimited` seems to imply that it's installing some clock that has its own behavior. "How do I control it?" would be a reasonable question, and the reasonable answer would be, "Well, install a `BroadcastFrameClock` instead."

So — counterproposal:

```kotlin
enum RecompositionClock {
  None, // rely on frame clock in calling context
  Immediate, // install eagerly recomposing clock
}
```

This also forces the caller to make a witting decision (which I think is a big improvement on the status quo), but has the advantage that it more clearly describes what's going on. Those who would actually want user-controlled frames or whatever can see that their path is to install a `BroadcastFrameClock` or whatever.

JakeWharton:
Hmm I don't think `None` quite makes sense because there _is_ always a clock. But I'm sympathetic to the case where it's a user-supplied, context-extracted clock vs. one we supply as being the choice.

I'll have to ruminate on it over night.

JakeWharton:
What do you think about `ExternalClock` or `ContextClock` for the enum value name? And its documentation would describe how the user-supplied `FrameClock` is used.

jingibus:
So like this?

```kotlin
enum RecompositionClock {
  /** 
   * Use the MonotonicFrameClock that already exists in the calling CoroutineContext. 
   * If none exists, an exception is thrown. 
   * 
   * Use this option to drive Molecule with the built-in Android frame clock.
   */
  ContextClock,
  /**
   * Install an eagerly recomposing clock. This clock will provide a new frame immediately whenever
   * one is requested. The resulting flow will emit a new item every time the snapshot state is invalidated.
   */
  Immediate,
}
```

That looks 💯 to me. I'll push those changes and we can nitpick from there

oldergod:
We don't run CI on this repo??
image

molecule/molecule-runtime/src/commonMain/kotlin/app/cash/molecule/GatedFrameClock.kt
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  private val clock = run {
+    var broadcastClock: BroadcastFrameClock? = null
+    broadcastClock = BroadcastFrameClock {
+      if (isRunning) sendFrame()
+    }
+    broadcastClock
+  }
```
JakeWharton:
```suggestion
  private val clock = BroadcastFrameClock {
    if (isRunning) sendFrame()
  }
```

JakeWharton (approved):
Looks good

jingibus:
Failures in the test module remind me that there *is* a test module. :) Do we want to follow up by deleting that?

JakeWharton:
Matt and I talked about it and thought it may be worth keeping for now for use with a context-based clock. But it's a little weird that it only really is useful in that mode. So long-term it's feasible we get rid of it and offer something a bit smaller like a `TestClock` which sends a frame and suspends until the frame is complete.

I would just try to make sure its tests pass in context-clock mode and not worry about immediate mode.

PR #95, opened 2022-07-25
Remove usage of unstable APIs

JakeWharton:
We do not want to break because someone updated their dependency but we also do not want to force the viral annotation either.

### DIFF
--- a/molecule/molecule-runtime/src/commonMain/kotlin/app/cash/molecule/GatedFrameClock.kt
+++ b/molecule/molecule-runtime/src/commonMain/kotlin/app/cash/molecule/GatedFrameClock.kt
@@ -21,8 +21,6 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.channels.Channel.Factory.CONFLATED
 import kotlinx.coroutines.launch
-import kotlin.time.ExperimentalTime
-import kotlin.time.TimeSource
 
 /**
  * A [MonotonicFrameClock] that is either running, or not.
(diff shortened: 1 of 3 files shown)


molecule/molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
```diff
@@ -123,7 +135,7 @@ fun <T> CoroutineScope.launchMolecule(
 ) {
   val clockContext = when (clock) {
     RecompositionClock.ContextClock -> EmptyCoroutineContext
-    RecompositionClock.Immediate -> GatedFrameClock(this).also { it.isRunning = true }
+    RecompositionClock.Immediate -> GatedFrameClock(this)
```
JakeWharton, line 138: This is the default and no `also` is/was used on line 60.

molecule/molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
```diff
@@ -76,10 +70,28 @@ private fun <T> immediateClockFlow(body: @Composable () -> T): Flow<T> = flow {
...
+    /*
+    TODO: Replace the above block with the following once `ReceiveChannel.isEmpty` is stable:
+
+    for (item in outputBuffer) {
+      emit(item)
+      if (outputBuffer.isEmpty) {
+        clock.isRunning = true
+      }
     }
+    */
```
JakeWharton, line 94: cc @jingibus

jingibus (approved):

molecule/molecule-runtime/src/commonMain/kotlin/app/cash/molecule/GatedFrameClock.kt
```diff
@@ -51,7 +47,7 @@ internal class GatedFrameClock(scope: CoroutineScope) : MonotonicFrameClock {
     }
 
   private fun sendFrame() {
-    clock.sendFrame(start.elapsedNow().inWholeNanoseconds)
+    clock.sendFrame(0L)
```
JakeWharton, line 50: We can add this (or expect/actual platform APIs for now) if someone complains, but I'm pretty sure no one needs/wants it.

PR #98, opened 2022-07-26
Deprecate testMolecule

JakeWharton:


molecule/molecule-testing/src/main/kotlin/app/cash/molecule/testing/moleculeTesting.kt
```diff
@@ -52,6 +52,11 @@ fun <T> testMolecule(
 ) = testMolecule(body, timeoutMs.milliseconds, validate)
 
 @ExperimentalCoroutinesApi
+@Deprecated(
+  """Use moleculeFlow or launchMolecule with an 'Immediate' clock and then use Turbine to test.
+
+If you believe this is insufficient please let us know at https://github.com/cashapp/molecule/issues/97.""",
```
JakeWharton, line 58: No `trimMargin()` in annotation literals :(

mattprecious: weak

molecule/molecule-testing/src/main/kotlin/app/cash/molecule/testing/moleculeTesting.kt
```diff
@@ -52,6 +52,11 @@ fun <T> testMolecule(
 ) = testMolecule(body, timeoutMs.milliseconds, validate)
 
 @ExperimentalCoroutinesApi
+@Deprecated(
+  """Use moleculeFlow or launchMolecule with an 'Immediate' clock and then use Turbine to test.
```
mattprecious, line 56: Is `launchMolecule` + `Turbine` a legitimate combination that can be made? Would be a BYO plumbing sort of deal?

mattprecious (approved):

molecule/molecule-testing/src/main/kotlin/app/cash/molecule/testing/moleculeTesting.kt
JakeWharton: Both APIs have the option of using the immediate clock. Since `StateFlow` is still a `Flow` you can still test it with Turbine.

mattprecious: Right. I forgot the signatures.

jingibus: I would recommend testing with `moleculeFlow` over `launchMolecule` due to [stateflow testing being annoying](https://www.billjings.net/posts/title/testing-stateflow-is-annoying/)

PR #104, opened 2022-08-11
Test GatedFrameClock properly

jingibus:
Followup for #101

### DIFF
--- a/molecule/molecule-runtime/src/commonJvmTest/kotlin/app/cash/molecule/MoleculeTest.kt
+++ b/molecule/molecule-runtime/src/commonJvmTest/kotlin/app/cash/molecule/MoleculeTest.kt
@@ -237,7 +237,7 @@ class MoleculeTest {
   @Test
   fun itemsImmediate() = runBlocking {
     val dispatcher = TestCoroutineDispatcher()
-    val values = Channel<Int>(Channel.UNLIMITED)
+    val values = Channel<Int>()
 
     val job = launch(dispatcher) {
       moleculeFlow(clock = RecompositionClock.Immediate) {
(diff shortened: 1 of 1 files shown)


molecule/molecule-runtime/src/commonJvmTest/kotlin/app/cash/molecule/MoleculeTest.kt
```diff
@@ -264,6 +264,14 @@ class MoleculeTest {
     value = values.awaitValue()
     assertEquals(2, value)
 
+    dispatcher.advanceTimeBy(100)
+    dispatcher.advanceTimeBy(100)
+    dispatcher.advanceTimeBy(100)
```
JakeWharton: Is this meaningfully different than 300? Or just being explicit?

JakeWharton (approved):

molecule/molecule-runtime/src/commonJvmTest/kotlin/app/cash/molecule/MoleculeTest.kt
jingibus: The goal is to bump against the backpressure mechanism. So:

1. Write into snapshot state (3), emitter called, bump against the rendezvous channel
2. Write into snapshot state (4), emitter not called
3. Write into snapshot state (5), emitter not called
4. Pull from channel, `moleculeFlow` calls `isRunning = true`, emitter called

So the answer to your question is "I have no idea"

jingibus: Real answer: Doesn't matter. Switched to one call with `300`, still fails. Revised and pushed

JakeWharton: Excellent. I fell asleep or I would have merged either way.

PR #109, opened 2022-08-30
Update tests to reflect testing with the Turbine library instead of t…

igorbzin:
…he deprecated molecule-test library

### DIFF
--- a/sample/build.gradle
+++ b/sample/build.gradle
@@ -9,7 +9,8 @@ dependencies {
   implementation libs.squareup.okhttp.logging.interceptor
 
   testImplementation libs.junit
-  testImplementation 'app.cash.molecule:molecule-testing'
+  testImplementation 'org.jetbrains.kotlinx:kotlinx-coroutines-test:1.6.4'
+  testImplementation 'app.cash.turbine:turbine:0.9.0'
 }
 
 android {
--- a/sample/build.gradle
+++ b/sample/build.gradle
@@ -9,8 +9,8 @@ dependencies {
   implementation libs.squareup.okhttp.logging.interceptor
 
   testImplementation libs.junit
-  testImplementation 'org.jetbrains.kotlinx:kotlinx-coroutines-test:1.6.4'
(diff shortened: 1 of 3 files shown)


sample/build.gradle
```diff
@@ -9,7 +9,8 @@ dependencies {
   implementation libs.squareup.okhttp.logging.interceptor
 
   testImplementation libs.junit
-  testImplementation 'app.cash.molecule:molecule-testing'
+  testImplementation 'org.jetbrains.kotlinx:kotlinx-coroutines-test:1.6.4'
+  testImplementation 'app.cash.turbine:turbine:0.9.0'
```
JakeWharton: Please use the version catalog for both of these dependencies.

JakeWharton (approved):

sample/build.gradle
igorbzin: Sure, I just added the change as well

igorbzin:
@JakeWharton I saw that some linter tasks failed, so I also changed the imports to be in lexicographical order.

I also saw that the build had other issues with incompatible Kotlin versions, but I am not sure where this is coming from:

`e: /home/runner/work/molecule/molecule/molecule/molecule-runtime/build/.transforms/31c98e32db1fd346b7dc7da7fe81e46b/transformed/out/jars/classes.jar!/META-INF/molecule-runtime_debug.kotlin_module: Module was compiled with an incompatible version of Kotlin. The binary version of its metadata is 1.7.1, expected version is 1.5.1.
[200](https://github.com/cashapp/molecule/runs/8103577171?check_suite_focus=true#step:6:201)`

JakeWharton:
Can you try bumping the Android Gradle plugin to 7.2.2? I can try here soon once I get to a computer.

igorbzin:
Bumping the AGP seems to have solved the Kotlin binary version errors, also it seems my Android Studio doesn't know alphabetical order, so I optimised imports again.

JakeWharton:
Thanks!

PR #113, opened 2022-09-13
Delete molecule-testing artifact

### LINKED ISSUE #97: Remove molecule-testing artifact
This is a tracking issue for removing the 'molecule-testing' artifact. You can test a Molecule with the `Immediate` clock and [Turbine](https://github.com/cashapp/turbine/).

Do you have a use case which is only covered by the 'molecule-testing' artifact and not Turbine? Let us know here.

JakeWharton:
Use Turbine.

Closes #97


molecule/kotlin-js-store/yarn.lock
```diff
@@ -55,6 +55,13 @@ brace-expansion@^1.1.7:
     balanced-match "^1.0.0"
     concat-map "0.0.1"
 
+brace-expansion@^2.0.1:
```
mattprecious, line 58: What's going on, here? Why are dependencies being bumped?

mattprecious (approved):
BYE

molecule/kotlin-js-store/yarn.lock
JakeWharton: Missed from the Kotlin 1.7 upgrade. Need to add a GitHub action that fails on dirty files in the repo after a build.

PR #131, opened 2022-11-07
Restructure to a (mostly) single project

JakeWharton:
We do some slight crimes to use the plugin that we're bulding in the build for this project iself.


mattprecious (approved):

oldergod:
Were there pain points with `includeBuilds` for even Molecule? Can still debug I presume?

JakeWharton:
Multiplatform projects don't really work with it.

oldergod:
Got it. Maybe that's why we're having so much trouble with Wire.

PR #132, opened 2022-11-08
Link thing

### LINKED ISSUE #120: Getting compiling error when using molecule in macos
`e: Module "moe.tlaster:precompose-molecule" has a reference to symbol app.cash.molecule/launchMolecule|-7305106088441339121[0]. Neither the module itself nor its dependencies contain such declaration.`

### LINKED ISSUE #124: Migrate tests to `commonTest`
We should be testing on all platforms

JakeWharton:
Closes #120. Closes #124.

Not sure what fixed the problem, but it fixed itself.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -21,6 +21,11 @@ jobs:
           distribution: 'zulu'
           java-version: 11
 
+      # For watchOS x86 simulator. Remove with Kotlin 1.8.
+      - uses: maxim-lobanov/setup-xcode@v1
+        with:
+          xcode-version: '13.4.1'
+
       - run: ./gradlew build assembleAndroidTest dokkaHtmlMultiModule
 
       - name: Run Tests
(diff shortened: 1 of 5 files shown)


mattprecious (approved):

.github/workflows/build.yaml
```diff
@@ -21,6 +21,11 @@ jobs:
           distribution: 'zulu'
           java-version: 11
 
+      # For watchOS x86 simulator. Remove with Kotlin 1.8.
+      - uses: maxim-lobanov/setup-xcode@v1
```
hfhbd, line 25: Do you have a youtrack ticket? I got the same error with Xcode 14.

JakeWharton: It's already fixed on their `master` so I didn't bother. The file is called `ExecutorService.kt` in their repo.

hfhbd: Okay, but thanks for the workaround! Hopefully, 1.8.0 beta will be out soon.

JakeWharton: Unfortunately it looks like it's currently only in Kotlin 1.8.20 according to GitHub, but perhaps it was cherry-picked 🤞

PR #133, opened 2022-11-09
Add watchos 32Bit support

hfhbd:


JakeWharton:
Do you actually need these? Or are they for completeness?

hfhbd:
Unfortunately, the S3 uses 32 bit and this processor is used in the Apple Watch 3, which was sold until September 2022, for stupid reasons... https://en.wikipedia.org/wiki/Apple_Watch#Support

hfhbd:
Any plans to publish these targets, eg in `0.6.1`?

JakeWharton:
Done

PR #135, opened 2022-11-16
Remove outdated sample

hfhbd:
I guess it is outdated. It is not used in the sample and is confusing.
(How do you emit Nothing at all?)

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,7 +61,7 @@ sealed interface ProfileModel {
 class ProfilePresenter(
   private val db: Db,
 ) {
-  fun transform(events: Flow<Nothing>): Flow<ProfileModel> {
+  fun transform(): Flow<ProfileModel> {
     return combine(
       db.users().onStart { emit(null) },
       db.balances().onStart { emit(0L) },


JakeWharton (approved):
This is fine.

All of our presenters are always in the form that accepts a single flow and returns a single flow. This is forced through the use of an interface. If you accept no inputs then `Flow<Nothing>` is the appropriate type because no events can come through. The way you create it is with a flow that simply never emits anything.

For the readme, though, there's no reason to bother since it's unused and we're not showing the interface or anything.

hfhbd:
Ah, you mean this https://code.cash.app/the-state-of-managing-state-with-compose?
Interesting, so you consume the inputs of the screen, don't you? Do you send the state of the whole screen "back" or do you create a single event for the changed input only? 
So if you have two textfields, do you use a sealed class with each class containing only the changed input of some properties, or do you simply use a data class and `copy(username = it)` or `copy(password = it)`?

```kt
@Composable
fun Login(models: StateFlow<LoginModel>, newEvent: (LoginEvent) -> Unit) {
  val model by models.collectAsState()
  TextField(model.username) { newEvent(UsernameChanged(it) }
  TextField(model.password) { newEvent(PasswordChanged(it) }
}

sealed interface LoginEvent {
data class UsernameChanged(val new: String): LoginEvent
data class PasswordChanged(val new: String): LoginEvent
```

Ideally, the sample could include some interaction too. I could add it but I don't know The Way.


In my playground app I used this implementation:
```kt
class LoginViewModel {
  private var username by mutableStateOf("")
     fun updateUsername(new: String) { username = new }

     private var password by mutableStateOf("")
     fun updatePassword(new: String) { password = new }

     fun state(
         coroutineScope: CoroutineScope,
         clock: RecompositionClock = RecompositionClock.ContextClock
     ): StateFlow<LoginState> = coroutineScope.launchMolecule(clock) { ... }
}

@Composable
 fun Login(viewModel: LoginViewModel) {
     val coroutineScope = rememberCoroutineScope()
     val state by remember(viewModel, coroutineScope) { viewModel.state(coroutineScope) }.collectAsState()

     Login(
         username = state.username,
         updateUsername = viewModel::updateUsername,
         password = state.password,
         updatePassword = viewModel::updatePassword,
         enableLogin = state.enableLogin,
         onLoginClick = viewModel::login,
         error = state.error
     )
 }

 @Composable
 private fun Login(
     username: String,
     updateUsername: (String) -> Unit,
     password: String,
     updatePassword: (String) -> Unit,
     enableLogin: Boolean,
     onLoginClick: () -> Unit,
     error: Failure?
 ) {
// real "normal" UI
}
```

JakeWharton:
We would not use Molecule for this level of granularity. Owning the state of the UI we leave entirely to the UI layer (whether it's written in Compose UI or views).

For a login screen there would probably only be a single event representing the action of logging in
```kotlin
data class SubmitEvent(val username: String, val password: String)
```
Maybe there's also a "forgot password" UI action which would get an event, too.

The state of the screen would probably be
```
sealed interface LoginModel
object Idle : LoginModel
object Loading : LoginModel
data class Failure(val errorMessage: String) : LoginModel
```
Depending on whether you have navigation side-effects happening from the UI layer or the presenter layer you may also add `object Success : LoginModel` to trigger navigation.

This design lets us test all three states of the UI layer with snapshot tests (idle, loading, failure) and it lets our presenter know nothing about how the UI layer is actually implemented.

hfhbd:
Thanks. So handling UI validation, e.g. username must contain 5 letters, you would move this into the UI layer? But doesn't this enforce you to duplicate the UI code for each target?
Is this the reason for redwood? :D

JakeWharton:
We don't share our presenters across platforms. We certainly could, in theory, and I've done it on other projects with similar architecture. The fact that you build the UI separately for each platform is a feature, not a bug. The UI design for a Chrome extension is very different than that of a command-line app which is very different than a mobile app or website.

Redwood is specifically about clients whose design is the same and for us, critically, can be updated outside of an app release. So the architectures are similar but don't necessarily have the same goals.

PR #137, opened 2022-11-24
Add foundation of sample that uses AAC ViewModel

chris-horner:
A common request I've heard from folks is an example of how to use Molecule to drive presentation logic when using androidx's `ViewModel`.

The plan here is to build a single screen sample app that hits [this API](https://dog.ceo/dog-api/) that displays random images of different dog breeds. I haven't actually implemented any of the actual logic yet, but that's not really the important part. What matters more is the skeleton proposed here.

### DIFF
--- a/sample-viewmodel/src/main/java/com/example/molecule/viewmodel/MoleculeViewModel.kt
+++ b/sample-viewmodel/src/main/java/com/example/molecule/viewmodel/MoleculeViewModel.kt
@@ -1,0 +1,50 @@
+/*
+ * Copyright (C) 2022 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.example.molecule.viewmodel
+
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.platform.AndroidUiDispatcher
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import app.cash.molecule.RecompositionClock.ContextClock
+import app.cash.molecule.launchMolecule
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.StateFlow
+
+abstract class MoleculeViewModel<Event, Model> : ViewModel() {
+  private val scope = CoroutineScope(viewModelScope.coroutineContext + AndroidUiDispatcher.Main)
+
+  // Events have a capacity large enough to handle simultaneous UI events, but
+  // small enough to surface issues if they get backed up for some reason.
+  private val events = MutableSharedFlow<Event>(extraBufferCapacity = 20)
+
+  val models: StateFlow<Model> by lazy(LazyThreadSafetyMode.NONE) {
+    scope.launchMolecule(clock = ContextClock) {
+      models(events)
+    }
+  }
+
+  fun take(event: Event) {
+    if (!events.tryEmit(event)) {
+      error("Event buffer overflow.")
+    }
+  }
+
+  @Composable
+  protected abstract fun models(events: Flow<Event>): Model
+}
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -14,6 +14,10 @@ squareup-retrofit = "2.9.0"
 android-plugin = { module = "com.android.tools.build:gradle", version = "7.3.1" }
 androidx-core = { module = "androidx.core:core-ktx", version = "1.9.0" }
 androidx-test-runner = { module = "androidx.test:runner", version = "1.5.2" }
+androidx-activity-compose = { module = "androidx.activity:activity-compose", version = "1.6.1" }
+
+androidx-compose-bom = { module = "androidx.compose:compose-bom", version = "2022.11.00" }
+androidx-compose-material3 = { module = "androidx.compose.material3:material3" }
 
 compose-compiler = { module = "org.jetbrains.compose.compiler:compiler", version.ref = "compose" }
 compose-runtime = { module = "org.jetbrains.compose.runtime:runtime", version = "1.2.2" }
--- a/sample-viewmodel/build.gradle
+++ b/sample-viewmodel/build.gradle
@@ -1,0 +1,9 @@
+apply plugin: 'com.android.application'
+apply plugin: 'org.jetbrains.kotlin.android'
+apply plugin: 'app.cash.molecule'
+
+dependencies {
+  implementation platform(libs.androidx.compose.bom)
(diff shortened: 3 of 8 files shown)


sample-viewmodel/src/main/java/com/example/molecule/viewmodel/Presenter.kt
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.example.molecule.viewmodel
+
+import androidx.compose.runtime.Composable
+import kotlinx.coroutines.flow.Flow
+
+interface Presenter<Event, Model> {
```
JakeWharton: I don't think people will want to do this and instead probably just want this function to be directly on `MoleculeViewModel`. Top-level, value-returning composables are also testable in isolation the same way instances of this class are.

JakeWharton (approved):
I guess this is fine? Honestly I have so little insight into the correct way to do anything with the Android recommended architecture because I find it so horrifically broken. I think most people writing `ViewModel` subclasses expose a bunch of state properties and bind them all individually so this still comes with a paradigm shift in that all of your state needs slammed together into a single model. Whether that's something they'll do in practice or not remains to be seen.

JakeWharton:
That's a general comment on Android's architecture recommendations and not your PR. Honestly if people migrate to this pattern then there's an easy off-ramp to get the hell out of using `ViewModel` completely which is a huge win.

JakeWharton:
Are we merging this as-is and working over multiple PRs?

chris-horner:
> I think most people writing ViewModel subclasses expose a bunch of state properties and bind them all individually

Yeah, I wanted to make it clear that the Android's `ViewModel` is glue, and the actual testable presentation logic should live elsewhere. We _could_ make this closer to how most folks use `ViewModel`, but I like the idea of demonstrating the advantages of not dumping presentation logic in a `ViewModel` subclass.

> Are we merging this as-is and working over multiple PRs?

If you're happy with this approach I'll stack some PRs on top of this one and we can merge a complete sample.

sample-viewmodel/src/main/java/com/example/molecule/viewmodel/MoleculeViewModel.kt
```diff
@@ -0,0 +1,49 @@
...
+    scope.launchMolecule(clock = ContextClock) {
+      models(events)
+    }
+  }
+
+  fun event(event: Event) {
+    if (!events.tryEmit(event)) {
+      error("Event buffer overflow.")
+    }
+  }
```
chris-horner: Realised I didn't handle event processing in my first attempt at this. Keep me honest here that this is OK 🙏

JakeWharton:
Yep, let's keep going.

sample-viewmodel/src/main/java/com/example/molecule/viewmodel/MoleculeViewModel.kt
```diff
@@ -0,0 +1,49 @@
...
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.viewModelScope
+import app.cash.molecule.RecompositionClock.ContextClock
+import app.cash.molecule.launchMolecule
+import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.MutableSharedFlow
+import kotlinx.coroutines.flow.StateFlow
+
+abstract class MoleculeViewModel<Event, Model> : ViewModel() {
```
jingibus, line 29: What if instead this:

```kotlin
override suspend fun show(state: MyVmApi.(MyUiState)->Unit) =
  launchMolecule {
    val uiState = remember { mutableStateOf(initialMyUiState) }
    val api = object : MyVmApi {
      // a bunch of methods that update uiState or do other stuff
    }

    // ??? maybe we don't need `key` here???
    key(uiState) {
      api.state(uiState)
    }
  }
```

jingibus: Could also preserve `uiState` by making it a property on `MoleculeViewModel`

sample-viewmodel/src/main/java/com/example/molecule/viewmodel/MoleculeViewModel.kt
```diff
@@ -0,0 +1,49 @@
...
+  private val scope = CoroutineScope(viewModelScope.coroutineContext + AndroidUiDispatcher.Main)
+  // Events have a capacity large enough to handle simultaneous UI events, but
+  // small enough to surface issues if they get backed up for some reason.
+  private val events = MutableSharedFlow<Event>(extraBufferCapacity = 20)
+
+  val models: StateFlow<Model> by lazy(LazyThreadSafetyMode.NONE) {
+    scope.launchMolecule(clock = ContextClock) {
+      models(events)
```
jingibus, line 38: Note that if we can get rid of `Flow` in our example here, that'd be awesome: `Flow` is useful plumbing, but nonoptimal where it can be avoided, as it adds a bit of buffering no matter what you do

sample-viewmodel/src/main/java/com/example/molecule/viewmodel/MoleculeViewModel.kt
chris-horner: OK, back from holiday and ready to look into this again. What'cha mean? Could you flesh out this approach? I'm having a hard time extrapolating what this would actually look like.

sample-viewmodel/src/main/java/com/example/molecule/viewmodel/MoleculeViewModel.kt
chris-horner: Like maybe keeping a `MutableState` that we update inside of the variant of `launchMolecule` that takes an emitter?

sample-viewmodel/src/main/java/com/example/molecule/viewmodel/MoleculeViewModel.kt
jingibus: Boy, it's been a while since I wrote this. But IIRC....

The idea is to get rid of `Flow` and have events just be methods. The implementations of those events just modify state. This is the "right" way to use a `ViewModel`, after all.

We can do that by defining the events as an API surface. So `MyVmApi` could also be called `MyEventSink` or something like that.

sample-viewmodel/src/main/java/com/example/molecule/viewmodel/MoleculeViewModel.kt
jingibus: Right. That's the end goal of the plumbing mentioned above. The implementation is less important than the idea.

chris-horner:
Haven't tried stacking PRs before on a forked repo, but heres the actual fleshed out implementation: https://github.com/chris-horner/molecule/pull/1

sample-viewmodel/src/main/java/com/example/molecule/viewmodel/MoleculeViewModel.kt
chris-horner: After toying around with it, I think that might be adding too much magic? The advantage of
```kotlin
@Composable
protected abstract fun models(events: Flow<Event>): Model
```
is that it's easy to reason about, and provides a simple API surface for producing an initial model and subsequent updates.

Happy to pair on this though if you have a clear vision in mind that you think would be better.

JakeWharton:
That looks great. Does that mean this is ready to merge?

jingibus:
I'm happy with merging it if y'all are

chris-horner:
👍

PR #164, opened 2023-01-10
Add implementation of ViewModel sample

chris-horner:
Not sure this needed to be quite so elaborate but I had fun with it. Was thinking of adding presenter tests in the next PR since this one is starting to get long, but lemme know. Happy to add them as part of this too.

https://user-images.githubusercontent.com/1245751/211670272-6c152cf2-8352-4487-8baf-8d8cea0c089f.mp4


JakeWharton (approved):

JakeWharton:
We should probably refactor the sample to look a little more like the ones in Redwood so that we can have the presenter Compose code in a module, the data layer in a module, and the two frontends (Compose UI+ViewModel w/Molecule, plain Compose UI w/o Molecule, View w/Molecule).

PR #169, opened 2023-01-17
Kotlin 1.8 and Compose compiler 1.4.0

### LINKED ISSUE #159: Kotlin 1.8-compatible release
Tracking issue

* [ ] Update JetBrains Compose to 1.8-compatible release

JakeWharton:
Closes #157. Closes #159.

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -2,10 +2,10 @@
 compileSdk = "33"
 minSdk = "21"
 
-compose = "1.3.2.2"
+composeCompiler = "1.4.0"
 coroutine = "1.6.4"
 
-kotlin = "1.7.20"
+kotlin = "1.8.0"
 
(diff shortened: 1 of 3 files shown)


mattprecious (approved):

JakeWharton:
Don't feel like dealing with the broken source set nonsense right now.

LouisCAD:
Is there a way to override the Compose Gradle plugin version that Molecule requests so we can update to Kotlin 1.8.10 or even to a Compose compiler pre-release/snapshot?

JakeWharton:
If you mean the Compose Kotlin compiler plugin then I would suggest using [Gradle's resolution strategy](https://docs.gradle.org/current/userguide/resolution_rules.html) to either force the version or use dependency substitution. It's added as a normal dependency to the plugin configuration for each target so the normal ways to manipulate a dependency will apply.

PR #170, opened 2023-01-21
Add DisplayLinkClock for iOS, tvOS, and macOS

kevincianfarini:
This commit adds two imlementations of a `MonotonicFrameClock`
backed by `CADisplayLink` on iOS and tvOS and `CVDisplayLink`
on macOS.

watchOS does not have an equivalent that I'm aware of and therefore
has been omitted.

### DIFF
--- a/molecule-runtime/build.gradle
+++ b/molecule-runtime/build.gradle
@@ -55,6 +55,66 @@ kotlin {
         implementation libs.androidx.core
       }
     }
+
+    darwinMain {
+      dependsOn(commonMain)
+    }
+
+    quartzCoreMain {
+      dependsOn(darwinMain)
+    }
+
+    iosArm64Main {
+      dependsOn(quartzCoreMain)
+    }
+
+    iosSimulatorArm64Main {
+      dependsOn(quartzCoreMain)
+    }
+
+    iosX64Main {
+      dependsOn(quartzCoreMain)
+    }
+
+    tvosArm64Main {
+      dependsOn(quartzCoreMain)
+    }
+
+    tvosSimulatorArm64Main {
+      dependsOn(quartzCoreMain)
+    }
+
+    tvosX64Main {
+      dependsOn(quartzCoreMain)
+    }
+
+    macosMain {
+      dependsOn(darwinMain)
+    }
+
+    macosArm64Main {
+      dependsOn(macosMain)
+    }
+
+    macosX64Main {
+      dependsOn(macosMain)
+    }
+
+    darwinTest {
+      dependsOn(commonTest)
+    }
+
+    macosTest {
+      dependsOn(darwinTest)
+    }
+
+    macosArm64Test {
+      dependsOn(macosTest)
+    }
+
+    macosX64Test {
+      dependsOn(macosTest)
+    }
   }
 }
 
--- a/molecule-runtime/build.gradle
+++ b/molecule-runtime/build.gradle
@@ -59,61 +59,26 @@ kotlin {
     darwinMain {
       dependsOn(commonMain)
     }
-
-    quartzCoreMain {
-      dependsOn(darwinMain)
-    }
-
-    iosArm64Main {
-      dependsOn(quartzCoreMain)
-    }
-
-    iosSimulatorArm64Main {
-      dependsOn(quartzCoreMain)
-    }
-
-    iosX64Main {
-      dependsOn(quartzCoreMain)
-    }
-
-    tvosArm64Main {
-      dependsOn(quartzCoreMain)
-    }
-
-    tvosSimulatorArm64Main {
-      dependsOn(quartzCoreMain)
-    }
-
-    tvosX64Main {
-      dependsOn(quartzCoreMain)
-    }
-
-    macosMain {
-      dependsOn(darwinMain)
-    }
-
-    macosArm64Main {
-      dependsOn(macosMain)
-    }
-
-    macosX64Main {
-      dependsOn(macosMain)
-    }
-
     darwinTest {
       dependsOn(commonTest)
     }
 
-    macosTest {
-      dependsOn(darwinTest)
+    quartzCoreMain {
+      dependsOn(darwinMain)
     }
-
-    macosArm64Test {
-      dependsOn(macosTest)
+    macosMain {
+      dependsOn(darwinMain)
     }
+  }
 
-    macosX64Test {
-      dependsOn(macosTest)
+  targets.each { target ->
+    if (target.name.startsWith('ios') || target.name.startsWith('tvos')) {
+      target.compilations.main.defaultSourceSet.dependsOn(sourceSets.quartzCoreMain)
+      // TODO Link against XCTest in order to get frame pulses on iOS/tvOS.
+      // target.compilations.test.defaultSourceSet.dependsOn(sourceSets.darwinTest)
+    } else if (target.name.startsWith('macos')) {
+      target.compilations.main.defaultSourceSet.dependsOn(sourceSets.macosMain)
+      target.compilations.test.defaultSourceSet.dependsOn(sourceSets.darwinTest)
     }
   }
 }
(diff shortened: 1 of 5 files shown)


molecule-runtime/build.gradle
```diff
@@ -55,6 +55,50 @@ kotlin {
...
+      dependsOn(darwinMain)
+    }
+
+    macosArm64Main {
+      dependsOn(macosMain)
+    }
+
+    macosX64Main {
+      dependsOn(macosMain)
+    }
```
kevincianfarini: I intend to make this more succinct before it merges. Notice the lack of tests -- this is something that needs to be smoke tested in a sample or integration test.

molecule-runtime/src/macosMain/kotlin/app/cash/molecule/DisplayLinkClock.kt
```diff
@@ -0,0 +1,54 @@
...
+import platform.CoreVideo.CVDisplayLinkStop
+import platform.CoreVideo.kCVReturnSuccess
+
+public actual object DisplayLinkClock : MonotonicFrameClock {
+
+  private val clock = BroadcastFrameClock {
+    // One or more awaiters have appeared. Start the DisplayLink clock callback so that awaiters
+    // get dispatched on the next available frame.
+    checkDisplayLink(CVDisplayLinkStart(displayLink.value))
+  }
```
kevincianfarini, line 38: I'm curious to hear opinions on this approach. Rather than making these clocks classes, I decided to make them singleton objects that only register their callbacks when at least one frame is awaiting dispatch.

molecule-runtime/src/quartzCoreMain/kotlin/app/cash/molecule/DisplayLinkClock.kt
```diff
@@ -0,0 +1,33 @@
...
+
+public actual object DisplayLinkClock : MonotonicFrameClock {
+
+  @Suppress("unused") // This registers a DisplayLink listener.
+  private val displayLink: CADisplayLink = CADisplayLink.displayLinkWithTarget(
+    target = this,
+    selector = NSSelectorFromString(this::tickClock.name),
+  )
+
+  private val clock = BroadcastFrameClock {
```
kevincianfarini, line 33: I'm also curious to hear if we think that `BroadcastFrameClock` is suitable for this. The documentation for it says the following: 

> This implementation is intended for low-contention environments involving low total numbers of threads in a pool on the order of ~number of CPU cores available for UI recomposition work, while avoiding additional allocation where possible.

Which makes me a bit uneasy. Thoughts?

molecule-runtime/src/quartzCoreMain/kotlin/app/cash/molecule/DisplayLinkClock.kt
```diff
@@ -0,0 +1,33 @@
...
+    // We only want to listen to the DisplayLink run loop if we have frame awaiters.
+    displayLink.addToRunLoop(NSRunLoop.currentRunLoop, NSRunLoop.currentRunLoop.currentMode)
+  }
+
+  override suspend fun <R> withFrameNanos(onFrame: (frameTimeNanos: Long) -> R): R {
+    return clock.withFrameNanos(onFrame)
+  }
+
+  @ObjCAction public fun tickClock() {
```
kevincianfarini, line 44: Apparently functions used by `NSSelectorFromString` need to be public or else they don't work. I haven't been able to confirm this because unit tests don't currently work.

molecule-runtime/src/macosMain/kotlin/app/cash/molecule/DisplayLinkClock.kt
JakeWharton: The Android one behaves the same way, although it's not a singleton.

kevincianfarini: Is there any benefit I might not be thinking for classes over a singleton here?

molecule-runtime/src/quartzCoreMain/kotlin/app/cash/molecule/DisplayLinkClock.kt
```diff
@@ -0,0 +1,33 @@
...
+    // Remove the DisplayLink from the run loop. It will get added again if new frame awaiters
+    // appear.
+    displayLink.removeFromRunLoop(NSRunLoop.currentRunLoop, NSRunLoop.currentRunLoop.currentMode)
+    clock.sendFrame(0L)
```
JakeWharton: The order of these statements differs across the implementations. Any reason?

molecule-runtime/src/quartzCoreMain/kotlin/app/cash/molecule/DisplayLinkClock.kt
JakeWharton: There's only one thread here. We aren't running concurrent recomposition.

molecule-runtime/src/quartzCoreMain/kotlin/app/cash/molecule/DisplayLinkClock.kt
kevincianfarini: Nope. I can make them consistent in my next commit when I re-add a macOS test and try to get it to pass.

molecule-runtime/src/quartzCoreMain/kotlin/app/cash/molecule/DisplayLinkClock.kt
kevincianfarini: This is probably better served as an inline comment than a github PR comment.

molecule-runtime/src/macosMain/kotlin/app/cash/molecule/DisplayLinkClock.kt
JakeWharton: Nope

molecule-runtime/src/quartzCoreMain/kotlin/app/cash/molecule/DisplayLinkClock.kt
JakeWharton: Makes sense since it's basically being used a callback from outside the class. With `@ObjCName` in Kotlin 1.8 we may be able to give it a stable name and have it still work when private, but I'm not sure I care all that much.

molecule-runtime/build.gradle
kevincianfarini: Got a macos test to pass; iOS and tvOS will require a full integration test. 

I'm struggling a bit to make this more succinct since this represents the following hierarchy. 

* darwinMain
  * quartzCoreMain
    * ios*Main (3 in total)
    * tvos*Main(3 in total)
  * macosMain
    * macosArmMain
    * macosX64Main

I could potentially build this as a list of lists and loop to configure, but that doesn't feel much better. Any tips for me?

kevincianfarini:
@JakeWharton this PR is ready for a real review now. One thing to note -- The current `darwinTest` sourceset only runs tests on macos targets. Adding the `quartzCoreTest` sourceset to `darwinTest` once the iOS simulator stuff is sorted should be all that's needed to test this.

JakeWharton:
To anyone watching this from the outside (because I keep linking it to people), we are going to merge, but are in the process of adding unit tests for iOS and tvOS.

swankjesse:
FYI, I borrowed some of this for Redwood.
https://github.com/cashapp/redwood/pull/990

molecule-runtime/src/quartzCoreMain/kotlin/app/cash/molecule/DisplayLinkClock.kt
```diff
@@ -0,0 +1,51 @@
...
+import platform.Foundation.NSSelectorFromString
+import platform.QuartzCore.CADisplayLink
+
+public actual object DisplayLinkClock : MonotonicFrameClock {
+
+  @Suppress("unused") // This registers a DisplayLink listener.
+  private val displayLink: CADisplayLink = CADisplayLink.displayLinkWithTarget(
+    target = this,
+    selector = NSSelectorFromString(this::tickClock.name),
+  )
```
kevincianfarini, line 31: I'm looking at this again months later and this doesn't need a variable. It can just be in an `init` block 🤦

kevincianfarini:
Woohoo! Thanks for getting this over the finish line, Jake.

PR #186, opened 2023-02-19
Add the rutime dependency to the project, not compilations

### LINKED ISSUE #136: IDE does not resolve dependency added by molecule plugin in Common source set
I have tried to include the molecule plugin in my KMM project, but even after applying the plugin, it seems that the dependency is not added properly to all sourceSets. I have done some further research and the issue very similar to this issue opened here:

https://youtrack.jetbrains.com/issue/KTIJ-18425/

with the difference, that it uses "implementation" instead of "compileOnly".

I opened a new issue here with a minimal project for reproducing attached:

https://youtrack.jetbrains.com/issue/KTIJ-23680/IDE-does-not-resolve-dependency-added-by-KotlinCompilerPluginSupportPlugin-in-Common-code

Until this is resolved, you will have to add 

               implementation("app.cash.molecule:molecule-runtime:0.6.0")

manually to the shared module build.gradle to make Molecule work in KMM projects.

JakeWharton:
This should resolve IDE issues in a multiplatform use.

Also fix the test which was wildly broken and add one for no supported plugin.

Closes #136


mattprecious (approved):
Why did `CountToThreeTest` get removed? Is this the test that you 'fixed'?

JakeWharton:
It was a test inside a test fixture. We already test the library normally. The fixture only is for guaranteeing the runtime is added automatically.

PR #229, opened 2023-05-06
Group Compose compiler into Kotlin update PRs

JakeWharton:
This works really well on paraphrase. Copying to here!

### DIFF
--- a/renovate.json5
+++ b/renovate.json5
@@ -9,7 +9,7 @@
       "groupName": "Kotlin and Compose",
       "matchPackagePrefixes": [
         "org.jetbrains.compose.compiler",
-        "org.jetbrains.kotlin",
+        "org.jetbrains.kotlin:kotlin",
       ],
     },
   ],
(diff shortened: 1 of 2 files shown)


renovate.json5
```diff
@@ -0,0 +1,16 @@
...
+  "extends": [
+    "config:base"
+  ],
+  "packageRules": [
+    {
+      // Compose compiler is tightly coupled to Kotlin version
+      "groupName": "Kotlin and Compose",
+      "matchPackagePrefixes": [
+        "org.jetbrains.compose.compiler",
+        "org.jetbrains.kotlin",
```
fardavide: That's a cool change! 👍

I noticed the project also has coroutines as a dependency; this 👇 would avoid grouping them together with Kotlin
```suggestion
        "org.jetbrains.kotlin:kotlin",
```

JakeWharton: Ah, right. I always forget it's a simple string prefix and not an actual groupId match.

mattprecious (approved):

Issue #237, opened 2023-05-23
Send real time in frame clocks, and test it


JakeWharton:
I don't remember if JS does, but the iOS/MacOS/tvOS ones definitely don't.

JakeWharton:
Can copy from https://github.com/cashapp/redwood/pull/1145

PR #261, opened 2023-07-06
Test that JS frame clock sends nanos

JakeWharton:
Refs #237

### DIFF
--- a/molecule-runtime/src/jsTest/kotlin/app/cash/molecule/WindowAnimationFrameClockTest.kt
+++ b/molecule-runtime/src/jsTest/kotlin/app/cash/molecule/WindowAnimationFrameClockTest.kt
@@ -15,18 +15,20 @@
  */
 package app.cash.molecule
 
+import assertk.all
+import assertk.assertThat
+import assertk.assertions.isLessThan
+import assertk.assertions.isPositive
 import kotlin.test.Test
-import kotlinx.coroutines.ExperimentalCoroutinesApi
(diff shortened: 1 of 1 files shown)


molecule-runtime/src/jsTest/kotlin/app/cash/molecule/WindowAnimationFrameClockTest.kt
```diff
@@ -15,18 +15,20 @@
...
-@OptIn(ExperimentalCoroutinesApi::class)
 class WindowAnimationFrameClockTest {
-  @Test fun ticks() = runTest {
-    val job = Job()
-    WindowAnimationFrameClock.withFrameNanos {
-      job.complete()
+  @Test fun ticksWithTime() = runTest {
+    val frameTimeA = WindowAnimationFrameClock.withFrameNanos { it }
+    val frameTimeB = WindowAnimationFrameClock.withFrameNanos { it }
+    assertThat(frameTimeA).all {
```
mattprecious, line 29: `all` is fun

mattprecious (approved):

molecule-runtime/src/jsTest/kotlin/app/cash/molecule/WindowAnimationFrameClockTest.kt
JakeWharton: Yep. It defers throwing until the end of the block and aggregates any failures. Very useful.

PR #263, opened 2023-07-07
Send real frame time in Darwin clocks

JakeWharton:
And test that it does.

Refs #237

### DIFF
--- a/molecule-runtime/src/darwinTest/kotlin/app/cash/molecule/DisplayLinkClockTest.kt
+++ b/molecule-runtime/src/darwinTest/kotlin/app/cash/molecule/DisplayLinkClockTest.kt
@@ -15,14 +15,20 @@
  */
 package app.cash.molecule
 
+import assertk.all
+import assertk.assertThat
+import assertk.assertions.isLessThan
+import assertk.assertions.isPositive
 import kotlin.test.Test
 import kotlinx.coroutines.test.runTest
 
 class DisplayLinkClockTest {
-
-  @Test fun `DisplayLinkClock delivers a single frame`() = runTest {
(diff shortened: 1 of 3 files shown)


JakeWharton:
Note that this test is disabled on iOS and tvOS as we have to run the test through XCTest and show a dummy application in order to get frame pulses.

molecule-runtime/src/darwinTest/kotlin/app/cash/molecule/DisplayLinkClockTest.kt
```diff
@@ -15,14 +15,20 @@
...
-
-  @Test fun `DisplayLinkClock delivers a single frame`() = runTest {
-    DisplayLinkClock.withFrameNanos {
-      // If this function does not time out the test passes.
+  @Test fun ticksWithTime() = runTest {
+    val frameTimeA = DisplayLinkClock.withFrameNanos { it }
+    val frameTimeB = DisplayLinkClock.withFrameNanos { it }
+    assertThat(frameTimeA).all {
+      isPositive()
+      isLessThan(frameTimeB)
```
oldergod, line 31: Is it testing that `frameTimeA` ran before `frameTimeB`? and we assume that the code is executed fast enough for it to make sense?

oldergod (approved):

molecule-runtime/src/darwinTest/kotlin/app/cash/molecule/DisplayLinkClockTest.kt
JakeWharton: Yes. Because it's nanos it's basically impossible to get two of the same values back since the precision is so high.

Issue #273, opened 2023-07-24
sample-viewmodel module showcases keeping an always hot StateFlow in the ViewModel


StylianosGakis:
This is an issue due to how AAC ViewModel typically is used. The backstack of screens keep the ViewModels in the backstack in memory, relying on behavior like `stateIn` and `collectAsStateWithLifecycle()` to turn the "hot" flows into "cold" ones when there no longer are observers on those states.

The example inside molecule could be altered to show a way which developers who still use ViewModels  can adopt in their own apps without this pretty important downside of keeping all molecule StateFlows always hot in the backstack.

This was briefly discussed already in https://github.com/cashapp/molecule/issues/271, but I do believe that it'd be worth it for the sample to be updated to solve this issue.

JakeWharton:
We are not users of ViewModel (the library), and would very much welcome any improvements to that sample which showcases the "right" way to use it with them.

StylianosGakis:
Yes, I totally understand this, and I really appreciate that you have a module like that in this repo in the first place. It really helped me for example take the first step to try out molecule, so thank you for that.

I've raised a PR here https://github.com/cashapp/molecule/pull/274. I'd love for someone else who's more familiar with AAC ViewModel to also take a look at it and discuss the approach I've taken to make those two libraries work with each other.

The one thing I am the least happy about is the fact that we lose the free synchronous first value that you get from using `launchMolecule`, but I don't see a way to make that work without having the StateFlow stay hot basically forever while it's in the backstack, as discussed here https://github.com/cashapp/molecule/issues/271#issuecomment-1644311575.

avelicu:
Hey friends,

I came across this library recently and I'm very excited to try it in my app but I consider this issue a blocker.

I have an alternative proposal PR https://github.com/cashapp/molecule/pull/572 that makes state observation more analogous to how it happens in the lifecycle-aware world on Android - providing explicit extension funs for collecting flows as state while the molecule is active, and repeating a coroutine while the molecule is active.

Note that the PR also provides a hammer approach to suspend the entire composition while the molecule is inactive, but I'm not sure anymore of the value of that approach.

I welcome all of your comments on the PR before I dive further into writing tests.

JakeWharton:
I think we should mostly abandon/rewrite this sample and focus on a solution using `retain { }` and Nav 3.

Issue #271, opened 2023-07-20
State production suspension and caching with Molecule


tunjid:
There are two questions in this:

1. suspending a launched Molecule: CoroutineScope.[launchMolecule](https://cashapp.github.io/molecule/docs/latest/molecule-runtime/app.cash.molecule/launch-molecule.html) implies in name and verifies in source that the launched Molecule will keep producing state while the launching `Coroutine` is active. In the case where there is no collector of the produced state, can there be a way to stop composing the `Presenter`?
2. If the above can be accommodated, can there be an API to seed the last produced state to presenters? This is so that resuming collecting from the presenter does not overwrite the last produced state value.

Consider the `ProfilePresenter` example:

```kotlin
@Composable
fun ProfilePresenter(
  userFlow: Flow<User>,
  balanceFlow: Flow<Long>,
): ProfileModel {
  val user by userFlow.collectAsState(null)
  val balance by balanceFlow.collectAsState(0L)

  return if (user == null) {
    Loading
  } else {
    Data(user.name, balance)
  }
}
```

The initial state is produced will always be `Loading`. In the case where `Data` has been produced and the `ProfilePresenter` is no longer being composed because the `StateFlow` is not being collected from, (maybe the owning screen has been placed in the back stack), I would like to prevent the last seen `Data` from immediately being overwritten by `Loading` as I return to the screen and the presenter produces state yet again.

i.e:
1. Screen in focus. State: Loading.
3. Data emitted. State: Data.
4. Screen in back stack, presenter is no longer composing. State: Data.
5. Screen back in focus, presenter is recomposed. Last seen Data state is overwritten. State: Loading.

The above can be worked around by using [`moleculeFlow`](https://cashapp.github.io/molecule/docs/latest/molecule-runtime/app.cash.molecule/molecule-flow.html) with the `Flow` builder to allow seeding as the `Flow` is cold, and then creating a `StateFlow` with the right `SharingStarted` argument:

```kotlin
class SeededStateHolder {
    var seed: ProfileModel = Loading
    val profileStateFlow = flow {
        // Pass the seed to the presenter to prevent overwriting state
        emitAll(moleculeFlow(ProfilePresenter(seed, ....))
    }
    // update the seed on each emission
    .onEach { seed = it }
    .stateIn(...)
}
```

Though I wonder if I'm missing something more obvious. If not, a built in API that allows for it would be really nice.

jingibus:
[...]

PR #274, opened 2023-07-24
Alter viewmodel-sample to show how to use a not always hot StateFlow

StylianosGakis:
Turn the StateFlow in MoleculeViewModel into one created using `stateIn` so that it can turn cold when there are no observers.

This PR is an effort to address https://github.com/cashapp/molecule/issues/273

Uses the ideas from here https://github.com/cashapp/molecule/issues/271#issue-1813467276 in order to solve the issue with when the ViewModel comes back from the backstack to still have the last emitted value available to it


sample-viewmodel/src/main/java/com/example/molecule/viewmodel/MoleculeViewModel.kt
```diff
@@ -45,6 +56,21 @@ abstract class MoleculeViewModel<Event, Model> : ViewModel() {
     }
   }
 
-  @Composable
-  protected abstract fun models(events: Flow<Event>): Model
```
tunjid, line 49: Consider amending this to

```kotlin
   abstract val initialSeed: Model

  @Composable
  protected abstract fun models(seed: Model, events: Flow<Event>): Model
``` 

Then to produce state in the `MoleculeViewModel`:

```kotlin
  private var seed = initialState

  val models: StateFlow<Model> by lazy(LazyThreadSafetyMode.NONE) {
    moleculeFlow(mode = ContextClock) {
      models(seed, events)
    }.onEach {
      seed = it
    }.stateIn(
      scope = scope,
      started = SharingStarted.WhileSubscribed(5.seconds),
      initialValue = seed,
    )
  }
```

With this change, the seed is explicitly needed in state production, and implementers of `MoleculeViewModel` have to provide an initial state.

When the presented `Flow` is collected from again, the last seen seed is provided so it can be restored.

StylianosGakis: A bit confused here, this part no longer is remembering the Presenter, should the burden of doing that then fall on the consumer instead?
Also, doing this
```
abstract val initialState: Model

private var seed: Model = initialState
```
Inside the `abstract class MoleculeViewModel` gives me a warning of "Accessing non-final property initialState in constructor" for the `seed`
image
I must be missing something here 👀

Tried this out here https://github.com/StylianosGakis/molecule/pull/1 and it seems like the warning I am getting does in fact make it so that when I run the app it finds `initialState` to be null because it's not initialized quick enough, so `seed` is also null, so the initial value inside the StateFlow is null, crashing the app.

Stacktrace starts like:
```
java.lang.NullPointerException: Parameter specified as non-null is null: method com.example.molecule.viewmodel.UiKt.PupperPicsScreen, parameter model
  at com.example.molecule.viewmodel.UiKt.PupperPicsScreen(Unknown Source:6)
```

tunjid: Ah sorry about that. The warning is right in that the parent class is dependent on the subclass for a value, the proper way to do this is to pass it in the constructor.

If you added an `initialState` argument to the `MoleculeViewmodel` constructor it should take care of the issue. This argument would then behave like the initial state in the `stateIn` operator on a `Flow`. The seed will be initialized to it and the rest should follow hopefully.

tunjid: I do think separating the presenter from the VM is a great pattern, my current suggestion optimizes for simplicity. Extend the VM, produce state.

Otherwise the presenter factory you have is great granted the returned presenter class is stateless. In your impl it's a class and it's not immediately obvious that on each restart a new instance of the class is created and any local state in the class lost. This is implicit in `models(seed, events) ` because there's no enclosing class body.

If you want to keep the factory, perhaps encourage that it's ought to be a stateless function by passing it as a constructor param too instead of a abstract method in a class. Then implement it with a static method that takes the repositories and the like instead of a class. 

```kotlin
abstract class MoleculeViewmodel<... >(
    InitialState: Model, 
    presenterFactory: (seed: Model, events: Flow<Event>) - >... 
) 
```

tunjid: So you'd have the `MoleculePresenter` defined as:

```kotlin
abstract class MoleculeViewModel<Event, Model>(
    InitialState: Model, 
    presenter: @Composable (seed: Model, events: Flow<Event>) - > Model
) : ViewModel() {
  ...
  private var seed = InitialState
  val models: StateFlow<Model> by lazy(LazyThreadSafetyMode.NONE) {
    moleculeFlow(mode = ContextClock) {
      presenter(seed, events)
    }.onEach {
      seed = it
    }.stateIn(
      scope = scope,
      started = SharingStarted.WhileSubscribed(5.seconds),
      initialValue = seed,
    )
  }
```

```kotlin
class PupperPicsViewModel(
    service: PupperService
) : MoleculeViewModel<Event, Model>(
    initialState = Model(
        loading = false,
        breeds = emptyList(),
        currentBreed = null,
        currentUrl = null,
      )
) { seed, events ->
    PupperPicsPresenter(seed, events, service)
}

@Composable
fun PupperPicsPresenter(seed: Model, events: Flow<Event>, service: PupperPicsService): Model {
  var breeds: List<String> by remember { mutableStateOf(seed.breeds) } 
  var currentBreed: String? by remember { mutableStateOf(seed.currentBreed) }
  var currentUrl: String? by remember { mutableStateOf(seed.currentUrl) }
  var fetchId: Int by remember { mutableStateOf(seed.fetchId) }

  ... return Model()
) 
```

In the above the `PupperPicsViewModel` provides the `PupperService`, pretty much your entry point to your DI graph, but the presenter is a static function you can test independently. You can change the seed and see how starting and restarting state production will go for example.

I'll clone your PR to validate my suggestions this time, lemme check the above compiles.

tunjid: Okay, I've validated the above. It works nicely bc the `PupperPicsPresenter` can be tested independently of the VM and the presenter test can be validated across platforms if you're building multiplatform.

 I have the following:

`MoleculeViewModel` definition:

```kotlin
abstract class MoleculeViewModel<Event, Model>(
  initialState: Model,
  started: SharingStarted = SharingStarted.WhileSubscribed(5.seconds),
  presenter: @Composable (seed: Model, events: Flow<Event>) -> Model,
) : ViewModel() {
  ...

  private var seed: Model = initialState

  val models: StateFlow<Model> by lazy(LazyThreadSafetyMode.NONE) {
    moleculeFlow(mode = ContextClock) {
      presenter(seed, events)
    }.onEach {
      seed = it
    }.stateIn(
      scope = scope,
      started = started,
      initialValue = seed,
    )
  }
  ...
}
```

```kotlin
class PupperPicsViewModel(
  // This service would typically be injected
  service: PupperPicsService = PupperPicsService(),
) : MoleculeViewModel<Event, Model>(
  initialState = Model(
    loading = false,
    breeds = emptyList(),
    currentBreed = null,
    currentUrl = null,
  ),
  presenter = { seed, events ->
    PupperPicsPresenter(
      seed = seed,
      events = events,
      service = service,
    )
  },
)
```

Then the `PupperPicsPresenter`:

```kotlin
@Composable
fun PupperPicsPresenter(
  seed: Model,
  events: Flow<Event>,
  service: PupperPicsService,
): Model {
  var breeds: List<String> by remember { mutableStateOf(seed.breeds) }
  var currentBreed: String? by remember { mutableStateOf(seed.currentBreed) }
  var currentUrl: String? by remember { mutableStateOf(seed.currentUrl) }
  var fetchId: Int by remember { mutableStateOf(0) }

  // state production invocations
  ...

   // return produced state
  return Model(
    loading = currentBreed == null,
    breeds = breeds,
    currentBreed = currentBreed,
    currentUrl = currentUrl,
  )
}
```

StylianosGakis: Passing it in the constructor works, thanks about that!
And yeah I totally see your comment about the factory approach where it's not obvious that it'll create a completely new instance of the class when it goes from cold to hot again.
I just wanted to try to mitigate the need to have to remember on the ViewModel impl, which might be easy to forget since you'd have to do it each time.

I suppose you could also do
```
class PupperPicsViewModel : MoleculeViewModel<Event, Model>(
  initialState = Model.InitialState,
) {
  private val presenter = PupperPicsPresenter(PupperPicsService())

  @Composable
  override fun models(seed: Model, events: Flow<Event>): Model {
    return presenter.present(seed, events)
  }
}
```
instead of:
```
class PupperPicsViewModel : MoleculeViewModel<Event, Model>(
  initialState = Model.InitialState,
) {
  @Composable
  override fun models(seed: Model, events: Flow<Event>): Model {
    val presenter = remember { PupperPicsPresenter(PupperPicsService()) }
    return presenter.present(seed, events)
  }
}
```
So it does stay in memory and not recreated as it was before.

The reason I was looking for an interface for the `MoleculePresenter` instead of a method, since you get the signature for free, you just do `class Foo : MoleculePresenter<Event, Model>` and you get the signature that you need instantly, don't need to remember to pass the Flow<Event> or forget to pass the `seed: Model`.

But at the same time you do get them in the `models` function in your VM, so it may be redundant too, not sure about that tradeoff 🤷‍♂️

StylianosGakis: I suppose at least for the sample here, the simpler solution should be chosen, meaning no MoleculePresenter interface but just a static function

StylianosGakis: Oh oops, for the past ~50 minutes I was testing things out and didn't see your last messages 🙈
But yeah you got it, that's perfect I think, and the lambda accepting the seed and the events flow feels so obvious in hindsight, can't believe I didn't think about it :D 

I still feel like passing the lambda or an interface is probably a decision for each team to make, but the more I see it the more I'd also lean towards your approach too.

Do you have access to push on top of my branch here? Maybe you can make a PR merging into mine here instead? I like your approach I think we can have that in the final PR for the molecule maintainers to also take a look at.

StylianosGakis: This PR has now been updated with the changes discussed here 👍

sample-viewmodel/src/test/java/com/example/molecule/viewmodel/PupperPicsPresenterTest.kt
```diff
@@ -30,19 +30,23 @@ import org.junit.Assert.assertEquals
...
+    loading = true,
+    breeds = emptyList(),
+    currentBreed = null,
+    currentUrl = null,
+  )
+
   @Test
   fun `on launch, breeds are loaded followed by an image url`() = runBlocking {
     val picsService = FakePicsService()
+    val presenter = PupperPicsPresenter(picsService)
```
tunjid: This change can probably be reverted since we can still write the `PupperPicsPresenter` inline as:

```kotlin
moleculeFlow(mode = RecompositionMode.Immediate) {
    PupperPicsPresenter(seed, emptyFlow(), picsService)
}
```

tunjid (approved):
LGTM from a ViewModel POV, deferring to molecule maintainers for if its still idiomatic molecule from their POV.

Looks like you may need to update some tests.

sample-viewmodel/src/test/java/com/example/molecule/viewmodel/PupperPicsPresenterTest.kt
StylianosGakis: Working on it 👌

StylianosGakis: Done

StylianosGakis:
Tests are fixed, thanks a lot for the help on this Tunji!

sample-viewmodel/src/main/java/com/example/molecule/viewmodel/presentationLogic.kt
```diff
@@ -31,22 +31,40 @@ sealed interface Event {
...
+  },
+)
+
 @Composable
-fun PupperPicsPresenter(events: Flow<Event>, service: PupperPicsService): Model {
-  var breeds: List<String> by remember { mutableStateOf(emptyList()) }
-  var currentBreed: String? by remember { mutableStateOf(null) }
-  var currentUrl: String? by remember { mutableStateOf(null) }
+fun PupperPicsPresenter(
+  seed: Model,
```
jingibus, line 61: This `seed` thing sure is awkward. There "should" be no need for it — as a practical matter, `moleculeFlow` will emit immediately. But obviously that is not something a `Flow` does...

sample-viewmodel/src/main/java/com/example/molecule/viewmodel/MoleculeViewModel.kt
```diff
@@ -20,31 +20,45 @@ import androidx.compose.ui.platform.AndroidUiDispatcher
...
-      models(events)
-    }
+    moleculeFlow(mode = ContextClock) {
+      presenter(seed, events)
+    }.onEach {
+      seed = it
+    }.stateIn(
+      scope = scope,
+      started = started,
+      initialValue = seed,
```
jingibus, line 55: The need for this `initialValue` does make the case that having a `started` parameter on `launchMolecule` is maybe not the worst idea in the world. Hooking it up with the existing API surface is uhh *waves hands* possible, for a certain definition of possible, but not one useful to someone who wants to do what `MoleculeViewModel` is doing

jingibus (approved):
This is a step up over the existing example, but `seed` here really does sell me on the merits of having a `started` parameter for `launchMolecule`, or some other solution that provides similar capabilties

sample-viewmodel/src/main/java/com/example/molecule/viewmodel/presentationLogic.kt
tunjid: The need for it mostly stems from restarting. Consider a timer that counts how much time is spent on a screen doing an activity, like the NYT crossword. It has the following presenter Composable:

```kotlin
@Composable
fun TimerPresenter(seed: Long): Long {
    var elapsed by remember { mutableStateOf(seed) }

    LaunchedEffect(Unit) {
        while(true) {
            delay(1.seconds)
            ++elapsed
       } 
    }
}
```

Without the seed, there is no way to restore elapsed to its last seen state in between presenter stop and restarts save for writing it to a data source. If the data source exposed a cold flow, the last seen value will immediately be overwritten:

```kotlin
@Composable
fun TimerPresenter(dataSource: Flow<Long>): Long {
    // There is no good default to use here upon resumption
    val savedElapsed by dataSource.collectAsState(0)
    // elapsed will start back at 0
    var elapsed by remember { mutableStateOf(savedElapsed) }

    LaunchedEffect(Unit) {
        while(true) {
            delay(1.seconds)
            ++elapsed
       } 
    }

    // Assuming the data source is well behaved and only emits once,
    // add the last time elapsed to the current one
    LaunchedEffect(savedElapsed) {
        elapsed += savedElapsed
    }
}
```

In the above, the UI will briefly flash `0` when the screen is returned to after state production is resumed after being stopped. The seed provides a way of "freezing" the last seen state for restoration when state production resumes as Compose Presenters unlike `Flow` require a synchronous (in as much as a Composable function can be deemed synchronous) initial return value.

tunjid: This isn't an issue with `Flow` operators like `combine` or `merge` because state won't be updated till all `Flows` emit in the case of the former, and till independent `Flows` emit in the latter. The immediate value requirement that Compose imposes seemingly mandates a seed value.

StylianosGakis:
I'd personally be happy to see this merged as it is right now, and for any thoughts that may arise around `seed` and if anything could be done on the library side can be addressed in a separate PR, after considering the discussion [here](https://github.com/cashapp/molecule/pull/274#discussion_r1274978690).

sample-viewmodel/src/main/java/com/example/molecule/viewmodel/presentationLogic.kt
Sripadmanabans: I don't understand the `TimerPresenter` example. It seems to only emit one value.
```kotlin
LaunchedEffect(Unit) {
    delay(1000L)
    ++elapsed
}
```
Should it be inside a `while (true)` loop?

tunjid: Yes, it should. Thanks for pointing it out!

Sripadmanabans: If we do end up using a `while (true)` in the presenter and convert the molecule flow with `stateIn` and  `SharingStarted.WhileSubscribed(5.seconds)` I think the behavior would be wrong. In the sense that we would be tracking an additional 5 seconds after the last collector has been removed.
Would it be better to use molecule to create a StateFlow that is always active and using events to start and stop the timer be a better way to keep track of screen time?

tunjid: In the particular case of this example, a no argument `SharingStarted.WhileSubscribed()` would have been used. The example is a bit trivial however, as other flows may also have been used in the presenter.

The larger point in context being, it would be nice if molecule presenters could be used to create a `StateFlow` that offers behaviors that match that in all permutations of `flow.stateIn(...)`. Especially in the case where the `Flow` is stopped and restarted, without overwriting the last emitted value.

sample-viewmodel/src/main/java/com/example/molecule/viewmodel/MoleculeViewModel.kt
JakeWharton, line 55: This whole pattern really feels like it should be encapsulated in something.

I could see it being one (or more) of the following:
- A valve. These existed for RxJava and basically let you control the upstream subscription based on another stream. In this case, the lifecycle would be filtered and mapped to a true/false stream based on whether the activity was resumed and that would control the valve on the upstream collection.
- Replaying share. Another RxJava thing, where an observable instance would cache the latest seen value, replay it to new subscribers, and unsubscribe upstream when the subscriber count dropped to 0.
- A custom frame clock (or frame clock wrapper) which only ticks when the lifecycle is resumed. The flow collector would always be present, but instead of using active collection to control whether recomposition occurs we use the frame clock.

I legitimately cannot figure out what to do with this PR. The use case is certainly valid. But the amount of code required to accomplish it as-is seems like it would overwhelm someone trying to learn how to use Molecule with AndroidX View Model rather than guide them.

JakeWharton: (This is not really a reflection of your PR, more of the current library design, the design of View Model, and the general sadness of lifecycle.)

StylianosGakis: > This is not really a reflection of your PR , more of the current library design, the design of View Model, and the general sadness of lifecycle.

Yes I understand and I agree. It really shouldn't be this way, but it is the current reality unfortunately.

> The use case is certainly valid. But the amount of code required to accomplish it as-is seems like it would overwhelm someone trying to learn how to use Molecule with AndroidX View Model rather than guide them.

So my entire reasoning behind starting this PR came down to the steps we took to adopt molecule in [the app](https://github.com/HedvigInsurance/android/blob/4cb58b21285b548e5541e9bc038d3875f855a3d7/app/molecule/molecule-android/src/main/kotlin/com/hedvig/android/molecule/android/MoleculeViewModel.kt#L19) I work for:

We were looking into adopting Molecule for our codebase for some time now.
1. A long time ago there was no viewmodel sample, and I looked briefly into it and I though "eh, it's probably gonna be hell to do in a ViewModel" so I dropped the idea.
2. Fast forward some time, the sample is there, and I look into it again. It all works using this tiny bit of glue code so I'm thinking let's go ahead and use it in this way exactly as it seems to work well.
3. I realize that the state flow is always hot, keeping potentially expensive flows hot on the backstack for all entries in the backstack which starts becoming a big deal. I pause this idea briefly
4. Then I tried to look for a way to make these two worlds play well together, hence this PR was made, where Tunji was very kind to help me through understanding how to work around ViewModel to make it happen.

Basically then, I fear that people who use AAC ViewModel will go up to step # 2 here, and then stop there without realizing there's more to be done. Then eventually when they figure out what happens, they'll blame Molecule instead of ViewModel since that's when this change would have been introduced to them.
So I thought if the sample played well with the lifecycle in the first place, all the people who would come here can see the easy and straightforward sample from the `sample` module here. And if they want a proper integration with ViewModel, they can look into the more involved `sample-viewmodel` module in this repo.

If we can do something to make this PR even better and make the code more understandable, perhaps from the points you suggest I'd love for that to happen of course. Other than that I personally still feel like it's a net positive if the sample which a lot of people are probably going to follow blindly is actually playing well with AAC ViewModel.

jingibus: @JakeWharton, if we consider `LaunchedEffect` this gets even gnarlier. Those are likely the pieces that you'd actually want to shut down — recomposition is designed not to have so many side effects.

On reflection, I think the happy places to land are likely to be:

* Include `rememberSaveable` support, then kill everything with a `stateIn` type solution, without any `seed`-type mechanism
* Just keep the molecule running, killing it when the ViewModel dies
* Include support for a `rememberRetained`, like Circuit has

The last two are most appealing to me. But even `rememberRetained` has weird consequences — IIRC one of the points of retention in ViewModel is to allow those long-running processes to keep going. The only approach we have that would achieve that would be to just keep it running.

And is keeping it running really so bad? How many would actually be kept running at one time in the VM framework? Is it just the VMs currently on screen during config change, or have I fallen off the boat here?

StylianosGakis: It's all the VMs in screens in the backstack, so potentially many of them.

tunjid: It's worth noting that a `rememberSaveable` solution would still be supsceptible to the overwrite on restart issue if the value to be restored is not parcelable.

The recommended period for state production to be halted in the absence observers with ViewModels is 5 seconds. If a screen is in the back stack for 5 seconds, it should still have access to it's last produced non parcelable state when resumed IMO. Otherwise, leaving a screen for > 5 seconds and returning to it will have the same behavior as the app cold starting on that screen.

In the case of a screen with a paginated list for example, this may mean a progress bar till the list is refreshed and the scroll position properly restored.

jingibus: > It's worth noting that a rememberSaveable solution would still be supsceptible to the overwrite on restart issue if the value to be restored is not parcelable.

Yep. `launchMolecule` should be capable of supporting all of `stateIn`'s parameters, which would fix that issue in a nice way.

Lucodivo:
If this change is not going to be merged, it seems only fair that the current sample-viewmodel contains a disclaimer of this known issue.

PR #283, opened 2023-07-28
Support custom version or coordinates for Compose compiler

### LINKED ISSUE #282: Provide the ability to set the JB Compose compiler version in our plugin
Same as https://github.com/cashapp/redwood/issues/1364

JakeWharton:
Since this project does not build a Kotlin compiler plugin, we should not be forced to do a release for new versions of Kotlin. This property will allow consumers to upgrade thier Kotlin version by specifying a newer JetBrains Compose compiler version, or a totally different set of Maven coordiantes for a custom Compose compiler aritfact.

Closes #282

### DIFF
--- a/molecule-gradle-plugin/src/main/java/app/cash/molecule/gradle/MoleculeExtension.kt
+++ b/molecule-gradle-plugin/src/main/java/app/cash/molecule/gradle/MoleculeExtension.kt
@@ -1,0 +1,40 @@
+/*
+ * Copyright (C) 2023 Square, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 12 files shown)


molecule-gradle-plugin/src/main/java/app/cash/molecule/gradle/MoleculeExtension.kt
```diff
@@ -0,0 +1,40 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.molecule.gradle
+
+import org.gradle.api.provider.Property
+
+interface MoleculeExtension {
+  /**
+   * The version of the JetBrains Compose compiler to use, or a Maven coordinate triple of
```
oldergod, line 22: Is "Maven coordinate triple" a thing?

oldergod (approved):

molecule-gradle-plugin/src/main/java/app/cash/molecule/gradle/MoleculeExtension.kt
JakeWharton: I've seen it referenced that way in a few places. It's meant to mean you need all three parts, whereas "coordinate" on its own may sometimes mean only the group and artifact IDs.

https://www.google.com/search?q=maven+coordinate+%22triple%22

PR #284, opened 2023-07-28
Ensure gated clock frame times do not collide on JS

### LINKED ISSUE #272: JS frame time can produce the same value
```
AssertionFailedError: expected to be less than:<57000000L> but was:<57000000L>
	at <global>.fail(/System/Volumes/Data/home/circleci/code/assertk/src/commonMain/kotlin/assertk/failure.kt:159)
	at <global>.expected(/System/Volumes/Data/home/circleci/code/assertk/src/commonMain/kotlin/assertk/assertions/support/support.kt:99)
	at <global>.isLessThan(/System/Volumes/Data/home/circleci/code/assertk/src/commonMain/kotlin/assertk/assertions/comparable.kt:25)
	at <global>.<unknown>(/Volumes/dev/cashapp/molecule/molecule-runtime/src/commonTest/kotlin/app/cash/molecule/GatedFrameClockTest.kt:33)
	at <global>.all(/System/Volumes/Data/home/circleci/code/assertk/src/commonMain/kotlin/assertk/failure.kt:181)
	at protoOf.doResume_5yljmg(/Volumes/dev/cashapp/molecule/molecule-runtime/src/commonTest/kotlin/app/cash/molecule/GatedFrameClockTest.kt:31)
	at protoOf.resumeWith_7onugl(/Volumes/dev/cashapp/molecule-runtime/build/compileSync/js/test/testDevelopmentExecutable/kotlin/commonMainSources/libraries/stdlib/src/kotlin/util/Standard.kt:55)
	at protoOf.resumeWith_s3a3yh(molecule-molecule-runtime-test.1609909856.js:47269)
	at protoOf.run_mw4iiu(/Volumes/dev/cashapp/molecule-runtime/build/compileSync/js/test/testDevelopmentExecutable/kotlin/commonMainSources/libraries/stdlib/src/kotlin/coroutines/Continuation.kt:45)
	at protoOf.processEvent_iukd42(/mnt/agent/work/44ec6e850d5c63f0/kotlinx-coroutines-test/common/src/TestDispatcher.kt:28)
```
Need to track the last seen frame time and +1 if it matches.

JakeWharton:
Closes #272

### DIFF
--- a/molecule-runtime/src/commonMain/kotlin/app/cash/molecule/GatedFrameClock.kt
+++ b/molecule-runtime/src/commonMain/kotlin/app/cash/molecule/GatedFrameClock.kt
@@ -46,8 +46,23 @@ internal class GatedFrameClock(scope: CoroutineScope) : MonotonicFrameClock {
       }
     }
 
+  private var lastNanos = 0L
+  private var lastOffset = 0
+
   private fun sendFrame() {
-    clock.sendFrame(nanoTime())
+    val timeNanos = nanoTime()
+
+    // Since we only have millisecond resolution, ensure the nanos form always increases by
+    // incrementing a nano offset if we collide with the previous timestamp.
+    val offset = if (timeNanos == lastNanos) {
+      lastOffset + 1
+    } else {
+      lastNanos = timeNanos
+      0
+    }
+    lastOffset = offset
+
+    clock.sendFrame(timeNanos + offset)
   }
 
   private val clock = BroadcastFrameClock {


JakeWharton:
I ran the test 1000 times and it didn't fail. Before it would fail about 10%.

molecule-runtime/src/commonMain/kotlin/app/cash/molecule/GatedFrameClock.kt
```diff
@@ -46,8 +46,23 @@ internal class GatedFrameClock(scope: CoroutineScope) : MonotonicFrameClock {
       }
     }
 
+  private var lastNanos = 0L
+  private var lastOffset = 0
```
mattprecious, line 50: Is there any reason to not just use a monotonically incrementing counter if we're going to +1 on a collision?

mattprecious (approved):

molecule-runtime/src/commonMain/kotlin/app/cash/molecule/GatedFrameClock.kt
JakeWharton: There's only 1,000,000 nanos in a millisecond but 4,000,000,000 ints. Actually 2,000,000,000 and then we roll over to -2,000,000,000. This basically means we skew the real time further and further until 2 whole seconds off and then we jump backwards to a -2 second offset (which also means you could observe the clock going backwards).

It sucks keeping both variables, but I can't see a path to anything simpler. I tried a bunch of things.

PR #285, opened 2023-08-01
Add APIs with exposed emitter

dzmpr:
## Problem:
We are use existing infrastructure of `rememberSaveable` to persist values in presenter, it requires to provide `SaveableStateRegistry` via `CompositionLocalProvider`. Current molecule API only takes composable body which return state. This API is not convenient for our use case.  `CompositionLocalProvider` can't be used because it do not return value from it's content lambda.
```kotlin
body = {
    CompositionLocalProvider(
        LocalSaveableStateRegistry provides saveableStateRegistry,
    ) {
        presenterComposable() // <- CLP do not return this value
    }
}
```
Now we use own implementation of CLP that can return value from content lambda:
```kotlin
@Composable
@OptIn(InternalComposeApi::class)
private inline fun withPresenterSaveableStateRegistry(
    content: @Composable () -> State,
): State {
    currentComposer.startProviders(
        values = arrayOf(LocalSaveableStateRegistry provides saveableStateRegistry),
    )
    val state = content()
    currentComposer.endProviders()
    return state
}
```
But this approach has at least two significant disadvantages:
1. We use internal compose API, that can change at every moment.
2. Our CLP recomposes at every composition.

## Proposal:
Molecule need alternative API with exposed emitter to be able to handle such cases:
```kotlin
body = { emitter ->
    CompositionLocalProvider(
        LocalSaveableStateRegistry provides saveableStateRegistry,
    ) {
        emitter(presenterComposable())
    }
}
```

Dunno is it possible to leave trailing lambda syntax for both bodies (with and w/o emitter). This head-on implementation has declaration clash when using trailing lambda. It requires to specify body parameter name to give a hint for compiler which lambda you want to use. Maybe it's possible to do this better, so this is not final code to approve.

### DIFF
--- a/molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
+++ b/molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
@@ -35,17 +35,36 @@ import kotlinx.coroutines.launch
 import kotlinx.coroutines.plus
 
 /**
+ * Create a [Flow] which will continually recompose `emittingBody` to produce a stream of [T] values
+ * when collected.
+ */
+public fun <T> moleculeFlow(
+  mode: RecompositionMode,
+  emittingBody: @Composable (Emitter<T>) -> Unit,
+): Flow<T> {
+  return when (mode) {
+    RecompositionMode.ContextClock -> contextClockFlow(emittingBody)
+    RecompositionMode.Immediate -> immediateClockFlow(emittingBody)
+  }
+}
+
+/**
  * Create a [Flow] which will continually recompose `body` to produce a stream of [T] values
  * when collected.
  */
 public fun <T> moleculeFlow(mode: RecompositionMode, body: @Composable () -> T): Flow<T> {
+  val emittingBody: @Composable (Emitter<T>) -> Unit = { emitter ->
+    emitter(body())
+  }
   return when (mode) {
-    RecompositionMode.ContextClock -> contextClockFlow(body)
-    RecompositionMode.Immediate -> immediateClockFlow(body)
+    RecompositionMode.ContextClock -> contextClockFlow(emittingBody)
+    RecompositionMode.Immediate -> immediateClockFlow(emittingBody)
   }
 }
 
-private fun <T> contextClockFlow(body: @Composable () -> T) = channelFlow {
+private fun <T> contextClockFlow(
+  body: @Composable (Emitter<T>) -> Unit,
+) = channelFlow {
   launchMolecule(
     mode = RecompositionMode.ContextClock,
     emitter = {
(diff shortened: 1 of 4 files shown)


JakeWharton:
> We use internal compose API, that can change at every moment.

These APIs cannot ever change without breaking every app in existence. They are internal because normal code should not be using them, but they are used inside inline functions and are thus part of the public API surface of Compose.

JakeWharton:
> Our CLP recomposes at every composition.

What's the cause? Is it due to the array creation inside the composable? Can you hoist that outside so its reference is stable and see if that helps?

Or is it simply due to the fact that Compose doesn't skip functions that return values?

dzmpr:
> Or is it simply due to the fact that Compose doesn't skip functions that return values?

Yes. To emit new state `presenterComposable` should be called, but due to it's inside 'our' CLP compose every time run provider code. Hoisting array is not an option.

JakeWharton:
> Hoisting array is not an option.

Why not? What about remembering then?

dzmpr:
> Why not? What about remembering then?

I mean recomposition of whole provider still happening when array hoisted.

molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
```diff
@@ -95,6 +116,32 @@ private fun <T> immediateClockFlow(body: @Composable () -> T): Flow<T> = flow {
...
+      if (outputFlow != null) {
+        outputFlow.value = value
+      } else {
+        flow = MutableStateFlow(value)
+      }
+    },
+    body = emittingBody,
+  )
+
+  return flow!!
```
JakeWharton, line 142: There is no guarantee the composition invoked the emitter so a `StateFlow` cannot be exposed. It has to be a `Flow`.

molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
```diff
@@ -130,8 +177,8 @@ public fun <T> CoroutineScope.launchMolecule(
  */
 public fun <T> CoroutineScope.launchMolecule(
   mode: RecompositionMode,
-  emitter: (value: T) -> Unit,
-  body: @Composable () -> T,
+  emitter: Emitter<T>,
+  body: @Composable (Emitter<T>) -> Unit,
```
JakeWharton, line 181: This would be a binary breaking change. In all likelihood this function would stay as-is and a new one would have to be added with a detailed description of when and why you would choose on vs. the other.

JakeWharton, line 181: It's not clear why this function even exists in this new form. Why pass the emitter to the function only to propagate it into the body to be invoked? The body can reference the emitter directly at the callsite–Molecule doesn't need to be involved.

This seems more like instead you want a function which sets up an applier-less Composition in which you can do whatever you want. In reality you can do that already today with Molecule by using `Unit` as your type and an empty `emitter` lambda.
```kotlin
val myEmitter: (String) -> Unit = ::println

launchMolecule(emitter = {}) {
  myEmitter("Hello!")
}
```

JakeWharton (commented):
The very first version of Molecule used a function call to emit values. While this offered a nice feature in that you could conditionally emit, it had so many other problems that it was deemed unworkable in practice. You can never call the emitter, you can call the emitter twice in a single composition, multiple "levels" of functions can invoke the emitter breaking composition.

I think due to the fact that you can already use `launchMolecule` with `Unit` as your type today and invoke whatever lambda you want means we don't have to actually ship any new APIs here. You can layer a helper function on top of that to expose whatever stream of values you want and leverage Molecule only for setting up and running the composition.

PR #312, opened 2023-10-04
Allow an optional CoroutineContext to be supplied to launchMolecule.

ghost:
Allow an optional CoroutineContext to be supplied to launchMolecule.

This allows easier control of what dispatcher Molecule should run under in cases where we can't presume it's wired correctly in the provided CoroutineScope.

### DIFF
--- a/molecule-runtime/api/android/molecule-runtime.api
+++ b/molecule-runtime/api/android/molecule-runtime.api
@@ -22,8 +22,12 @@ public final class app/cash/molecule/AndroidUiFrameClock : androidx/compose/runt
 }
 
 public final class app/cash/molecule/MoleculeKt {
-	public static final fun launchMolecule (Lkotlinx/coroutines/CoroutineScope;Lapp/cash/molecule/RecompositionMode;Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function2;)V
-	public static final fun launchMolecule (Lkotlinx/coroutines/CoroutineScope;Lapp/cash/molecule/RecompositionMode;Lkotlin/jvm/functions/Function2;)Lkotlinx/coroutines/flow/StateFlow;
+	public static final fun launchMolecule (Lkotlinx/coroutines/CoroutineScope;Lapp/cash/molecule/RecompositionMode;Lkotlin/coroutines/CoroutineContext;Lkotlin/jvm/functions/Function2;)Lkotlinx/coroutines/flow/StateFlow;
+	public static final fun launchMolecule (Lkotlinx/coroutines/CoroutineScope;Lapp/cash/molecule/RecompositionMode;Lkotlin/jvm/functions/Function1;Lkotlin/coroutines/CoroutineContext;Lkotlin/jvm/functions/Function2;)V
+	public static final synthetic fun launchMolecule (Lkotlinx/coroutines/CoroutineScope;Lapp/cash/molecule/RecompositionMode;Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function2;)V
+	public static final synthetic fun launchMolecule (Lkotlinx/coroutines/CoroutineScope;Lapp/cash/molecule/RecompositionMode;Lkotlin/jvm/functions/Function2;)Lkotlinx/coroutines/flow/StateFlow;
+	public static synthetic fun launchMolecule$default (Lkotlinx/coroutines/CoroutineScope;Lapp/cash/molecule/RecompositionMode;Lkotlin/coroutines/CoroutineContext;Lkotlin/jvm/functions/Function2;ILjava/lang/Object;)Lkotlinx/coroutines/flow/StateFlow;
+	public static synthetic fun launchMolecule$default (Lkotlinx/coroutines/CoroutineScope;Lapp/cash/molecule/RecompositionMode;Lkotlin/jvm/functions/Function1;Lkotlin/coroutines/CoroutineContext;Lkotlin/jvm/functions/Function2;ILjava/lang/Object;)V
 	public static final fun moleculeFlow (Lapp/cash/molecule/RecompositionMode;Lkotlin/jvm/functions/Function2;)Lkotlinx/coroutines/flow/Flow;
 }
 
--- a/molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
+++ b/molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
@@ -101,12 +101,14 @@ private fun <T> immediateClockFlow(body: @Composable () -> T): Flow<T> = flow {
  * to produce a [StateFlow] stream of [T] values.
  */
 public fun <T> CoroutineScope.launchMolecule(
+  context: CoroutineContext = EmptyCoroutineContext,
   mode: RecompositionMode,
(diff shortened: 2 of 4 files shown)


molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
```diff
@@ -101,12 +101,14 @@ private fun <T> immediateClockFlow(body: @Composable () -> T): Flow<T> = flow {
  * to produce a [StateFlow] stream of [T] values.
  */
 public fun <T> CoroutineScope.launchMolecule(
+  context: CoroutineContext = EmptyCoroutineContext,
   mode: RecompositionMode,
   body: @Composable () -> T,
 ): StateFlow<T> {
```
JakeWharton: This and the function below need to be proper overloads rather than just adding a new parameter. We already shipped 1.0 so this would break binary compatibility.

I'll add the binary compatibility ~validator~ dumper plugin tomorrow.

molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
```diff
@@ -143,7 +146,10 @@ public fun <T> CoroutineScope.launchMolecule(
     val recomposer = Recomposer(coroutineContext)
     val composition = Composition(UnitApplier, recomposer)
     var snapshotHandle: ObserverHandle? = null
-    launch(start = UNDISPATCHED) {
+    launch(
+      context = context,
+      start = UNDISPATCHED
```
JakeWharton:
```suggestion
      start = UNDISPATCHED,
```
I assume ktlint will ask for this too

JakeWharton (approved):
Can we write a test? You can add an arbitrary key to the context and assert it's present in the composition for something that works on all platforms.

molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
waylon-brown: Sorry, I was having some issues getting the gradle build happy with my JDK, addressed [here](https://github.com/cashapp/molecule/pull/312/commits/b5092be63556903e67dc6cc0e8438e56d2aae7a8) (though it looks like this also formatted the whole test file).

molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
waylon-brown: Good call, I would have seen this if I had the tests set up from the get-go. Let me know if adding telescoping [here](https://github.com/cashapp/molecule/pull/312/commits/b5092be63556903e67dc6cc0e8438e56d2aae7a8) addressed the binary compatibility.

waylon-brown:
> Can we write a test? You can add an arbitrary key to the context and assert it's present in the composition for something that works on all platforms.

Added a [test](https://github.com/cashapp/molecule/pull/312/commits/b5092be63556903e67dc6cc0e8438e56d2aae7a8), and this also required an additional change [here](https://github.com/cashapp/molecule/pull/312/commits/052166d98b105423539281597ca6bf44a07b35bb) since the test made me aware I wasn't using the passed in context for the Recomposer.

oldergod:
Just in case, you might want to rebase on master for the apiDump task would fail otherwise after merge

ghost:
> Just in case, you might want to rebase on master for the apiDump task would fail otherwise after merge

Rebased onto master.

molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
```diff
@@ -133,17 +151,43 @@ public fun <T> CoroutineScope.launchMolecule(
...
+  body: @Composable () -> T,
 ) {
   val clockContext = when (mode) {
     RecompositionMode.ContextClock -> EmptyCoroutineContext
     RecompositionMode.Immediate -> GatedFrameClock(this)
   }
 
   with(this + clockContext) {
-    val recomposer = Recomposer(coroutineContext)
+    val recomposer = Recomposer(context)
```
ghost: **Question to the reviewers:** Before we merge this, I want to make sure we're doing the right thing here. On re-reviewing, I'm wondering if the Recomposer now has lost context of the `clockContext` just above. Should the change instead here be

```
with(this + clockContext + context)
```
and remove each of the usages of `context` below (and switch back to `coroutineContext` here)?

JakeWharton: Yes, the scope context needs combined with the external context which needs amended with the clock context. Clock should be last since we want the immediate clock to take precedence over any clock registered in the external context.

JakeWharton: If you don't get to it, I'll take over the PR tomorrow and make the necessary changes (along with some style nits) and land it. Mostly out today.

ghost: Thanks, [updated!](https://github.com/cashapp/molecule/pull/312/commits/e9b89457226d881c5403ac4921f059994cf77026)

molecule-runtime/api/android/molecule-runtime.api
```diff
@@ -22,8 +22,12 @@ public final class app/cash/molecule/AndroidUiFrameClock : androidx/compose/runt
...
-	public static final fun launchMolecule (Lkotlinx/coroutines/CoroutineScope;Lapp/cash/molecule/RecompositionMode;Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function2;)V
-	public static final fun launchMolecule (Lkotlinx/coroutines/CoroutineScope;Lapp/cash/molecule/RecompositionMode;Lkotlin/jvm/functions/Function2;)Lkotlinx/coroutines/flow/StateFlow;
+	public static final fun launchMolecule (Lkotlinx/coroutines/CoroutineScope;Lapp/cash/molecule/RecompositionMode;Lkotlin/coroutines/CoroutineContext;Lkotlin/jvm/functions/Function2;)Lkotlinx/coroutines/flow/StateFlow;
+	public static final fun launchMolecule (Lkotlinx/coroutines/CoroutineScope;Lapp/cash/molecule/RecompositionMode;Lkotlin/jvm/functions/Function1;Lkotlin/coroutines/CoroutineContext;Lkotlin/jvm/functions/Function2;)V
+	public static final synthetic fun launchMolecule (Lkotlinx/coroutines/CoroutineScope;Lapp/cash/molecule/RecompositionMode;Lkotlin/jvm/functions/Function1;Lkotlin/jvm/functions/Function2;)V
+	public static final synthetic fun launchMolecule (Lkotlinx/coroutines/CoroutineScope;Lapp/cash/molecule/RecompositionMode;Lkotlin/jvm/functions/Function2;)Lkotlinx/coroutines/flow/StateFlow;
```
JakeWharton, line 28: This diff and the tool that produces it are really subpar, but these two lines are the same as the ones which were deleted except with `synthetic` added (i.e., hiding them solely for binary compatibility). The new source-compatible versions are the two lines above and the synthetic functions for supporting their default argument values are the two lines below.

PR #314, opened 2023-10-05
Add Kotlin JVM binary dump tool

JakeWharton:
It says that it does validation, but that's a lie. Future changes still require care to ensure compatibility.


oldergod (approved):
But at least, it tells you if something has changed without your knowing.

PR #322, opened 2023-10-23
Renovate: Use stable Ktlint coordinates

hfhbd:
Ktlint did change their group name since 1.0.0


JakeWharton:
With ktlint becoming stable I believe we can remove this configuration altogether.

hfhbd:
Ktlint is stable, so let's wait for a spotless update using ktlint 1.0.0?

JakeWharton:
I believe the version we're on defaults to 1.0. It was added in 6.22.0.

PR #359, opened 2024-01-08
Crossinline all the lambdas that reasonably can be.

kyay10:
This is a very minor change. I understand that exposing all these functions in ABI might not be reasonable. If so, ignore this PR.


JakeWharton:
Is there a motivating reason or goal here? Something measured or observed?

PR #392, opened 2024-02-29
Add PR template

JakeWharton:


mattprecious (approved):
Is this something that we want to require of external contributors?

JakeWharton:
Yep

JakeWharton:
It took me an hour to release Redwood 0.8.0 because I had to trawl through the commits from 4 months to make the change log. It took me like 5 minutes to release 0.9.0 because we had adopted this and started enforcing it. So now I will be doing it basically everywhere.

PR #513, opened 2024-09-23
Fix nano calculation

JakeWharton:
https://github.com/JakeWharton/mosaic/issues/481

---


JakeWharton:
Whoops guess we don't block on reviews here

oldergod (commented):
LGTM

Issue #544, opened 2024-11-26
`MoleculeTest.errorInEmitterDelayed` flaky


JakeWharton:
```
> Task :molecule-runtime:macosArm64Test

app.cash.molecule.MoleculeTest.errorInEmitterDelayed[macosArm64] FAILED
    kotlin.NoSuchElementException at /Users/runner/work/molecule/molecule/molecule-runtime/src/commonTest/kotlin/app/cash/molecule/MoleculeTest.kt:213
```

FletchMcKee:
### Cause
After adding some extensive logging, I discovered that the error is not with `errorInEmitterDelayed` but instead the `errorImmediately` and `errorInEmitterImmediately` tests. The flakiness in these tests appears to be caused by cancellation leaking between tests, I'm guessing due to shared coroutine scopes using BroadcastFrameClock. 

I noticed that when I added a log inside the catch block of `recomposer.runRecomposeAndApplyChanges()` inside `launchMolecule()` the cancellation log from the immediately tests would leak into the delayed tests:
```kotlin
public fun <T> CoroutineScope.launchMolecule(
  mode: RecompositionMode,
  emitter: (value: T) -> Unit,
  context: CoroutineContext = EmptyCoroutineContext,
  body: @Composable () -> T,
) {
  ..
  launch(finalContext, start = UNDISPATCHED) {
    try {
      recomposer.runRecomposeAndApplyChanges()
    } catch (e: CancellationException) {
      println("runRecomposeAndApplyChanges() cause: ${e.cause}")
      composition.dispose()
      snapshotHandle?.dispose()
    }
  }
  ..
}

// MoleculeTest.kt
@Test fun errorImmediately() {
   ..
  scope.cancel("errorImmediately")
}
```

<img width="850" src="https://github.com/user-attachments/assets/818712be-9153-4902-96af-d54206f73aae" />
<p/>

The flakiness came down to a timing issue:
- If the leaked cancellationException occurred before the first `registerGlobalWriteObserver` it went unnoticed. From what I saw, this appeared to always be the case with the android tests. 
- However with the macosArm64 and iosSimulatorArm64 platforms, what I would see is it would occur after `registerGlobalWriteObserver` and the cancellation would swallow the pending recomposition inside `runRecomposeAndApplyChanges()`, preventing the simulated runtimeException from being thown.

### Solution
Updating the immediately tests calling `scope.cancel()` with using UnconfinedTestDispatcher() appears to have solved the issue for me locally:
[...]

PR #581, opened 2025-02-17
MoleculeTest.errorInEmitterDelayed flaky

FletchMcKee:
This PR is an effort to address the issue in [#544](https://github.com/cashapp/molecule/issues/544)

### Cause
After adding some extensive logging, I discovered that the error is not with `errorInEmitterDelayed` but instead the `errorImmediately` and `errorInEmitterImmediately` tests. The flakiness in these tests appears to be caused by cancellation leaking between tests. 

I noticed that when I added a log inside the catch block of `recomposer.runRecomposeAndApplyChanges()` inside `launchMolecule()` the cancellation log from the immediately tests would leak into the delayed tests:
```kotlin
  mode: RecompositionMode,
  emitter: (value: T) -> Unit,
  context: CoroutineContext = EmptyCoroutineContext,
  body: @Composable () -> T,
) {
  ..
  launch(finalContext, start = UNDISPATCHED) {
    try {
      recomposer.runRecomposeAndApplyChanges()
    } catch (e: CancellationException) {
      println("runRecomposeAndApplyChanges() cause: ${e.cause}")
      composition.dispose()
      snapshotHandle?.dispose()
    }
  }
  ..
}

// MoleculeTest.kt
@Test fun errorImmediately() {
   ..
  scope.cancel("errorImmediately")
}
```

<p/>

The flakiness came down to a timing issue:
- If the leaked cancellationException occurred before the first `registerGlobalWriteObserver` it went unnoticed. From what I saw, this appeared to always be the case with the android tests. 
- However with the macosArm64 and iosSimulatorArm64 platforms, what I would see is it would occur after `registerGlobalWriteObserver` and the cancellation would swallow the pending recomposition inside `runRecomposeAndApplyChanges()`, preventing the simulated runtimeException from being thown.

### Solution
Updating the immediately tests calling `scope.cancel()` with using UnconfinedTestDispatcher() appears to have solved the issue for me locally. Can't say with complete certainty this solves the issue as it was difficult reproducing the error without any changes, but it does consistently contain the scope cancellation within those tests. Alternatively you could add `runTest` to these and use its coroutineContext + job + clock for the CoroutineScope, but that maybe defeats the purpose of the immediately context of these tests.

---

### DIFF
--- a/molecule-runtime/src/commonTest/kotlin/app/cash/molecule/MoleculeStateFlowTest.kt
+++ b/molecule-runtime/src/commonTest/kotlin/app/cash/molecule/MoleculeStateFlowTest.kt
@@ -33,14 +33,13 @@ import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.CoroutineStart
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.Job
-import kotlinx.coroutines.cancel
+import kotlinx.coroutines.cancelAndJoin
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.job
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.plus
 import kotlinx.coroutines.supervisorScope
-import kotlinx.coroutines.test.UnconfinedTestDispatcher
 import kotlinx.coroutines.test.advanceTimeBy
 import kotlinx.coroutines.test.runCurrent
 import kotlinx.coroutines.test.runTest
@@ -87,9 +86,10 @@ class MoleculeStateFlowTest {
     job.cancel()
   }
 
-  @Test fun errorImmediately() {
+  @Test fun errorImmediately() = runTest {
+    val job = Job()
     val clock = BroadcastFrameClock()
-    val scope = CoroutineScope(UnconfinedTestDispatcher() + clock)
+    val scope = CoroutineScope(coroutineContext + job + clock)
 
     // Use a custom subtype to prevent coroutines from breaking referential equality.
     val runtimeException = object : RuntimeException() {}
@@ -99,7 +99,7 @@ class MoleculeStateFlowTest {
       }
     }.isSameInstanceAs(runtimeException)
 
-    scope.cancel()
+    job.cancelAndJoin()
   }
 
   @Test fun errorDelayed() = runTest {
--- a/molecule-runtime/src/commonTest/kotlin/app/cash/molecule/MoleculeStateFlowTest.kt
+++ b/molecule-runtime/src/commonTest/kotlin/app/cash/molecule/MoleculeStateFlowTest.kt
@@ -40,6 +40,7 @@ import kotlinx.coroutines.job
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.plus
 import kotlinx.coroutines.supervisorScope
+import kotlinx.coroutines.test.UnconfinedTestDispatcher
 import kotlinx.coroutines.test.advanceTimeBy
 import kotlinx.coroutines.test.runCurrent
 import kotlinx.coroutines.test.runTest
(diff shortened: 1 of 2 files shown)


JakeWharton:
Thanks for such a thorough investigation! My only concern here is that we're somewhat playing whack-a-mole with which tests we're updating, rather than solving the problem completely.

It sounds like we should be doing a `cancelAndJoin` on the scope, perhaps? This will wait for all of the child jobs to cancel and return before allowing the test to complete. This should prevent the condition that you describe is causing the flakiness, I think.

FletchMcKee:
No problem! And that's a valid point, I also know UnconfinedTestDispatcher can be a common source of flaky tests. I was actually initially going to recommend changing those `error...Immediately` tests to replicate what most of the other tests are doing by using `runTest` and adding `coroutineContext + job` to the scope. Cancelling the job in this manner also fixed this issue, but then I realized the tests ending with `Immediately` were possibly omitting the `runTest` on purpose. That's where I figured UnconfinedTestDispatcher could make sense in this context.

However this appears to work nicely and avoids adding a TestDispatcher, so I can definitely update with this change:
```kotlin
@Test fun errorImmediately() = runTest {
  val clock = BroadcastFrameClock()
  val scope = CoroutineScope(clock)

  // Use a custom subtype to prevent coroutines from breaking referential equality.
  val runtimeException = object : RuntimeException() {}
  assertFailure {
    scope.launchMolecule(ContextClock, emitter = { fail() }) {
      throw runtimeException
    }
  }.isSameInstanceAs(runtimeException)

  scope.coroutineContext[Job]?.cancelAndJoin()
}
```

FletchMcKee:
Looked into this a bit more last night and you were right to say this change may not address the real problem completely, even if it prevents flaky tests. I discovered the compositions and snapshotHandles in `errorDelayed`, `errorDelayedImmediate` and `errorInEmitterDelayed` are continuing to run after their tests have completed. Their scope is cancelled, but it isn't propagating to the try/catch here in the `launchMolecule` method:
```kotlin
  val recomposer = Recomposer(finalContext)
  launch(finalContext, start = UNDISPATCHED) {
    try {
      recomposer.runRecomposeAndApplyChanges()
    } catch (e: CancellationException) {
      composition.dispose()
      snapshotHandle?.dispose()
    }
  }
```
So `composition` and `snapshotHandle` are never being disposed. This is different from other tests using a delay like `errorInEffect` which is triggering this CancellationException block and disposing these fields. ~~Not quite sure why this is different other than maybe the LaunchedEffect is cancelling the parent coroutineContext internally.~~[^1]

I noticed these tests are still producing logs in the `Snapshot.registerGlobalWriteObserver` ObserverHandle when other tests are running, but it essentially does nothing because the scope is already cancelled so the `Snapshot.sendApplyNotifications()` never gets called. This could explain why flaky tests aren't occurring with these compositions leaking into other tests vs the `error...Immediately` tests as the scope in those tests was still active.

The `job.cancel()` at the end of these delayed tests is already cancelled before this is invoked. ~~I'm not familiar enough with these Snapshot and Recomposer classes to understand why throwing a RuntimeException immediately allows the scope to remain active but throwing it delayed cancels the scope but isn't triggering the cancellationException~~[^2], which keeps the composition and snapshotHandle active. However I believe this is at the crux of this issue. 

I did verify that adding `composition.dispose()` and `snapshotHandle?.dispose()` in a `try/finally` or `invokeOnCompletion` block prevents the compositions from running after their tests have completed, but this could change the nature of how `launchMolecule` is expected to work. Thought it was at least worth sharing this information, but not sure what the best path forward would be.

**Edit:**
[^1]: LaunchedEffect does appear to cancel the scope [internally](https://github.com/androidx/androidx/blob/androidx-main/compose/runtime/runtime/src/commonMain/kotlin/androidx/compose/runtime/Effects.kt#L288).
[^2]: The `assertFailure` block is catching the exception, so it is never propagated to the scope. This is why the `error...Immediately` scopes continued to run and cause flakiness. The `error...Delayed` test exceptions are thrown so the scopes are cancelled. Unlike the LaunchedEffect test cases, it appears a cancellationException is never propagated which is why the snapshotHandle and composition are not disposed.

FletchMcKee:
Created a more robust PR

PR #596, opened 2025-03-13
Fix errorDelayed flaky tests

### LINKED ISSUE #544: `MoleculeTest.errorInEmitterDelayed` flaky

FletchMcKee:
This should be a more robust solution for the [flaky test issues](https://github.com/cashapp/molecule/issues/544) that have occurred from these `errorDelayed` and `errorInEmitterDelayed` tests. 

I understand if changes to the molecule.kt class are beyond the scope of this issue, but I did observe scenarios where the coroutine launching `recomposer.runRecomposeAndApplyChanges()` would complete without throwing a CancellationException, and this would result in leaking the `snapshotHandle`. However the changes in the test class should remain flaky-free whether this change is accepted or not.

Early on I was fairly confident as to _what_ was causing this issue, which was coroutine/snapshotHandle leaking between tests. Whenever there's a flaky test situation where the success rate is this high, it's a fair assumption. But I wasn't able to figure _why_ and _how_ this was causing this issue until recently after combing through the Snapshot and Recomposer classes.

There were two things that would always occur in a failure result that would never occur in a successful result, and that was `clock.hasAwaiters` would be false and `Snapshot.isApplyObserverNotificationPending` would be true at this moment before attempting to call `clock.sendFrame(0)`
```kotlin
@Test fun errorDelayed() = runTest {
 ..
  count++
  runCurrent()
  if (!clock.hasAwaiters) {
    // Test is guaranteed to fail and `Snapshot.isApplyObserverNotificationPending` is always true
  }
  clock.sendFrame(0)
  runCurrent()
  ..
}
```

When I updated the `errorImmediately` and `errorInEmitterImmediately` tests to using `runTest` and added `job.cancelAndJoin()` at the end, I was unable to reproduce the test failures, but it wasn't adding up as to why `errorImmediately` would affect the result in `errorDelayed` when the tests don't share a coroutineScope or recomposer. The only thing they could share is a Snapshot. After going through the Snapshot and Recomposer classes, I think I might finally know why this was happening, and although the window is small, it isn't improbable. Here is the sequence of events that cause the error:

1. The `error_Immediately` snapshotHandle is still undisposed when the `error_Delayed` test begins.
    
    -  This was almost always the case even in successful tests, however the test ordering may vary by platform, and the issue is more likely to occur if a delayed test directly follows an immediate test.
2. The `error_Immediately` is cancelled and its snapshotHandle disposal begins after `count++`, intercepting the `advanceGlobalSnapshot` with modified changes.
    
    - This is another thing that can occur in successful test cases, but is always part of the failure process. 
    - When the disposal begins, it removes the global write observer and calls `advanceGlobalSnapshot`. Without `error_Delayed` updating the count, no modified changes would occur. However in this scenario, the `error_Delayed` creates the change (count++), but `error_Immediately` wins the race when it comes to who advances the snapshot with modified changes. I knew these methods were potentially significant as it's the only place where `pendingApplyObserverCount.add` occurs and this is what causes `Snapshot.isApplyObserverNotificationPending` to be true:
```kotlin
fun registerGlobalWriteObserver(observer: ((Any) -> Unit)): ObserverHandle {
  ..
  return ObserverHandle {
    sync {
      globalWriteObservers -= observer
    }
    advanceGlobalSnapshot()
  }
}

private fun <T> advanceGlobalSnapshot(block: (invalid: SnapshotIdSet) -> T): T {
  ..
  // Critical point
  modified?.let {
    try {
      val observers = applyObservers
      observers.fastForEach { observer ->
        observer(it.wrapIntoSet(), previousGlobalSnapshot)
      }
    } finally {
      pendingApplyObserverCount.add(-1)
    }
  }
  ..
}
```
    
3. The `error_Delayed`'s `clock.sendFrame(0)` occurs _before_ these changes are applied to the observers.
   
     - It took me a while to figure out why, but then I realized all of the `runCurrent()` calls in these tests have no effect since the Snapshot is being advanced from `error_Immediately`'s disposal. The `Snapshot.sendApplyNotifications()` calls are occurring, whether through the external call in the test or through the internal call inside `launchMolecule`, but they have no effect because this `changes` boolean is false at this point since the immediately test has taken a new snapshot (but not updated the observers yet)
```kotlin
fun sendApplyNotifications() {
  val changes = sync {
    currentGlobalSnapshot.get().modified?.isNotEmpty() == true
  }
  if (changes)
    advanceGlobalSnapshot()
}
```

I probably went way too in-depth there, but I'm fairly confident this is what is happening. If the immediately tests were using the same scheduler, the `runCurrent()` calls would complete this process and allow the recomposer to record the modification and tick the `withFrameNanos`, but because it isn't using a scheduler, the timing will be fairly random. I've observed scenarios where the immediately tests advance the global snapshot and succeed, but it's because it completes before that critical `sendFrame(0)` part.

I'm still skeptical about changing the try/catch to try/finally in your `launchMolecule` method as your team will be far more knowledgeable on how this should run, but through my testing I would see the `error_Delayed` tests' snapshotHandles still being notified of updates. This may not be something observed in production and more related to how `runTest` can hang when you have uncompleted coroutines. I can't think of a scenario where `runRecomposeAndApplyChanges()` would complete without a CancellationException and you would want to keep the composition and snapshotHandle around, but that scenario may certainly exist.

Closes #544

---

### DIFF
--- a/molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
+++ b/molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
@@ -24,7 +24,6 @@ import androidx.compose.runtime.snapshots.Snapshot
 import kotlin.DeprecationLevel.HIDDEN
 import kotlin.coroutines.CoroutineContext
 import kotlin.coroutines.EmptyCoroutineContext
-import kotlin.coroutines.cancellation.CancellationException
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.CoroutineStart.UNDISPATCHED
 import kotlinx.coroutines.channels.Channel
@@ -180,7 +179,7 @@ public fun <T> CoroutineScope.launchMolecule(
   launch(finalContext, start = UNDISPATCHED) {
     try {
       recomposer.runRecomposeAndApplyChanges()
-    } catch (e: CancellationException) {
+    } finally {
       composition.dispose()
       snapshotHandle?.dispose()
     }
--- a/molecule-runtime/src/commonTest/kotlin/app/cash/molecule/MoleculeStateFlowTest.kt
+++ b/molecule-runtime/src/commonTest/kotlin/app/cash/molecule/MoleculeStateFlowTest.kt
@@ -21,19 +21,19 @@ import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableIntStateOf
 import androidx.compose.runtime.remember
 import androidx.compose.runtime.setValue
-import androidx.compose.runtime.snapshots.Snapshot
 import app.cash.molecule.RecompositionMode.ContextClock
 import app.cash.molecule.RecompositionMode.Immediate
 import assertk.assertFailure
 import assertk.assertThat
 import assertk.assertions.isEqualTo
 import assertk.assertions.isSameInstanceAs
+import assertk.assertions.isTrue
 import kotlin.test.Test
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.CoroutineStart
 import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.Job
-import kotlinx.coroutines.cancel
+import kotlinx.coroutines.cancelAndJoin
 import kotlinx.coroutines.delay
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.job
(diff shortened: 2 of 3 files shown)


JakeWharton (approved):

JakeWharton:
Thanks!

FletchMcKee:
No problem, glad I could help!. For whatever reason, I enjoy working on flaky test situations.

PR #605, opened 2025-03-30
Removing moshi and kSP.

### LINKED ISSUE #603: Remove KSP+Moshi, use kotlinx.serialization
Don't want a KSP dep if we don't need it.

Sripadmanabans:
Replacing moshi and ksp with kotlin serialization.

Fixes #603.
---


JakeWharton (approved):
Wow thanks!

PR #610, opened 2025-04-09
Add enum to control whether we register a snapshot notifier

### LINKED ISSUE #542: Add `SnapshotPolicy` argument to functions
This controls whether or not the `Composer`/`Recomposer` setup also registers a global write observer that sends `Snapshot` apply notifications. If you have another Compose-based toolkit in your app (e.g., Compose UI) that is already sending these, there is no reason to run both.

Let's add this with #382

JakeWharton:
Closes #542

---


molecule-runtime/src/commonMain/kotlin/app/cash/molecule/SnapshotNotifier.kt
```diff
@@ -0,0 +1,59 @@
...
+   * snapshot writes and sending apply notifications. Usually this means that some other
+   * Compose-based system is being used in your application, and that it will always be
+   * initialized prior to Molecule or at the same time.
+   *
+   * Failure to ensure someone else is sending apply notifications will result in state writes
+   * not triggering additional recomposition.
+   *
+   * Some examples where this policy can be used:
+   * - On Android, using Compose UI _before_ Molecule (e.g., calling `setContent { }` on an
+   *   `Activity` or `ComposeView`) will create a global snapshot write listener and applier.
```
mattprecious: This is a sticky state, right? E.g., if I create and attach a `ComposeView` and then detach it, I will still have a listener/applier wired up?

molecule-runtime/src/commonMain/kotlin/app/cash/molecule/SnapshotNotifier.kt
```diff
@@ -0,0 +1,59 @@
...
+   *   launches is valid.
+   * - On JetBrains' Compose UI for Desktop, calling the `application { }` function is enough to
+   *   ensure their global snapshot write listener and applier is started (you do not have to even
+   *   show a window).
+   */
+  External,
+
+  /**
+   * Register a global snapshot write observer and send apply notifications when new writes occur
+   * via a coroutine launched on the same scope as the molecule was launched. This coroutine will
```
mattprecious:
```suggestion
   * via a coroutine launched on the same scope as the Molecule was launched. This coroutine will
```

Not sure if you're being deliberate about proper noun usages vs not.

mattprecious (approved):

molecule-runtime/src/commonMain/kotlin/app/cash/molecule/SnapshotNotifier.kt
JakeWharton: I've been trying to avoid it, actually. I don't think we call them "molecules" anywhere else.

molecule-runtime/src/commonMain/kotlin/app/cash/molecule/SnapshotNotifier.kt
JakeWharton: Yes. I can add a word to indicate that, like maybe "singleton" or something.

PR #615, opened 2025-04-11
Add system property to control snapshot notifier default

JakeWharton:
Since the Compose UI snapshot notifier is a global singleton that can be started in one place, this allows you to also switch Molecule's default in that same location. Much easier than updating tens or hundreds of call sites.

---


molecule-runtime/src/javaTest/kotlin/app/cash/molecule/DefaultSnapshotNotifierTest.kt
```diff
@@ -0,0 +1,58 @@
...
+import assertk.assertions.isEqualTo
+import kotlin.test.Test
+
+// Note: We do not share this constant with the production code to verify its value doesn't change.
+private const val property = "app.cash.molecule.snapshotNotifier"
+
+class DefaultSnapshotNotifierTest {
+  @Test fun noProperty() {
+    assertThat(defaultSnapshotNotifier()).isEqualTo(WhileActive)
+  }
```
mattprecious: Is it possible to share this test with non-JVM? There's a (small) risk of drift otherwise.

molecule-runtime/src/javaMain/kotlin/app/cash/molecule/platform.kt
```diff
@@ -0,0 +1,24 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package app.cash.molecule
+
+internal actual fun defaultSnapshotNotifier(): SnapshotNotifier {
+  return System.getProperty("app.cash.molecule.snapshotNotifier")
+    ?.let { property ->
+      runCatching { SnapshotNotifier.valueOf(property) }.getOrNull()
```
mattprecious, line 21: Any thoughts about how we would safely make migrations to this enum in the future that don't lead to runtime crashes when people update? Just rely on existing binary compatibility practices/tests?

mattprecious (approved):

molecule-runtime/src/javaTest/kotlin/app/cash/molecule/DefaultSnapshotNotifierTest.kt
JakeWharton: Moved to common. Technically every other test implicitly validates this as well. I'm also not testing that the default value is being used, really.

molecule-runtime/src/javaMain/kotlin/app/cash/molecule/platform.kt
JakeWharton: Yeah, you really can't. They're set in stone forever.

Issue #613, opened 2025-04-10
Some test is flaky and hangs


JakeWharton:
See https://github.com/cashapp/molecule/actions/runs/14384606655/attempts/1, for example.

Not the first time I've seen it, either.

JakeWharton:
Pretty confident it's `MoleculeConcurrentTest` ☹

FletchMcKee:
I'm fairly certain it is the MoleculeConcurrentTest class that is causing the hanging builds. You can reproduce it locally by placing the test in a loop:

```kotlin
@Test fun coroutineContextHonoredByImmediateClock() = repeat(2) { /* ... existing test ... */ }
```

I believe the problem is that the original test relies on recomposition occurring naturally, but the implicit StandardTestDispatcher used by `runTest` doesn't drive Dispatchers.Default—which is where the recomposition is scheduled. Since nothing is forcing recomposition to occur, the `cancelLatch.await()` hangs indefinitely.

Compose runtime has a similar test in their [RecomposerTests.jvm.kt](https://github.com/androidx/androidx/blob/androidx-main/compose/runtime/runtime/src/nonEmulatorJvmTest/kotlin/androidx/compose/runtime/RecomposerTests.jvm.kt#L55-L86) class. You could modify your `coroutineContextHonoredByImmediateClock` test to do something similar:
```kotlin
@Test fun coroutineContextHonoredByImmediateClock() = runTest(UnconfinedTestDispatcher()) {
  val testThread = Thread.currentThread()

  val job = Job()
  val threadChannel = Channel<Thread>(Channel.UNLIMITED)
  lateinit var recomposeScope: RecomposeScope
  backgroundScope.launchMolecule(Immediate, job + Dispatchers.Default) {
    threadChannel.trySend(Thread.currentThread())
    val scope = currentRecomposeScope
    SideEffect { recomposeScope = scope }
  }

  val firstThread = threadChannel.receive()
  assertThat(firstThread).isSameInstanceAs(testThread)

  recomposeScope.invalidate()

  val secondThread = threadChannel.receive()
  assertThat(secondThread).isNotSameInstanceAs(testThread)
  assertThat(secondThread.name).contains("DefaultDispatcher")

  job.cancelAndJoin()
}
```

cmck0655:
I can reopen #617 again since it looks like there are still issues with hanging builds.

FletchMcKee:
I can reopen #617 since it looks like there are still hanging builds.

[...]

PR #617, opened 2025-04-12
Fix for hanging MoleculeConcurrentTest

FletchMcKee:
This should hopefully resolve the issue #613 for the hanging test. I believe this should achieve the desired results of verifying the Immediate clock starts on the current thread without the use of backward writes.

The existing test case had the risk of recomposing on every frame as it was writing state that had already been read within the composable. If UnconfinedTestDispatcher is undesirable, I can look into alternatives for making sure the recompositon runs predictably.

---


FletchMcKee:
Closing due to inactivity

PR #666, opened 2025-08-21
Add some notes to README about Compose vs Compose UI, and that Molecule is not a framework

zach-klippenstein:
---


README.md
```diff
@@ -17,6 +17,8 @@ fun CoroutineScope.launchCounter(): StateFlow<Int> = launchMolecule(mode = Conte
 }
 ```
```
JakeWharton, line 19: Stealing from https://github.com/JakeWharton/mosaic/?tab=readme-ov-file#i-thought-jetpack-compose-was-a-ui-toolkit-for-android a bit
```suggestion

Wait–I thought Jetpack Compose was a UI toolkit for Android?

Compose is, at its core, a general-purpose compiler and runtime to do state tracking and tree node and property manipulation.
This can be used on any platform supported by Kotlin for any kind of state or with any tree.
It's an amazing piece of technology.

Molecule "just" glues Compose's state management to kotlinx.coroutines' flows so that it can be used without the node tree.

```

JakeWharton (approved):

JakeWharton (approved):

PR #688, opened 2025-10-17
Use new `webMain` source set and HMPP 

JakeWharton:
New Dokka doesn't allow shared source folders.

---


oldergod (approved):
🍠

PR #698, opened 2025-11-17
Replaced material icons with ImageVector

FletchMcKee:
- Added the rounded_arrow_drop_down_24.xml directly from the Resource Manager now that material icons is no longer bundled with the material3 library.

---


JakeWharton:
Woo, thanks!

PR #730, opened 2026-02-07
  Fix flaky MoleculeConcurrentTest by replacing runTest with runBlocking

### LINKED ISSUE #613: Some test is flaky and hangs

Amir-yazdanmanesh:
The root cause is that `runTest` uses `StandardTestDispatcher` which has a dispatch timeout mechanism. The test's `cancelLatch.await()` suspends the test body waiting for a
   recomposition on real `Dispatchers.Default` threads. Since `StandardTestDispatcher` has no pending tasks (all work is on `Dispatchers.Default`), `runTest` may time out or
  hang.

  `runBlocking` is the correct choice because:
  - The test uses real `Dispatchers.Default` (not virtual time)
  - `runBlocking` naturally blocks until async work completes with no artificial dispatch timeout
  - The test's structure (blocking on a `CompletableDeferred`) maps directly to `runBlocking` semantics

  ## Test plan
  - [x] `./gradlew :molecule-runtime:jvmTest --tests "app.cash.molecule.MoleculeConcurrentTest" --rerun` passes consistently (verified 5 consecutive runs)

  Fixes #613

### DIFF
--- a/molecule-runtime/src/jvmTest/kotlin/app/cash/molecule/MoleculeConcurrentTest.kt
+++ b/molecule-runtime/src/jvmTest/kotlin/app/cash/molecule/MoleculeConcurrentTest.kt
@@ -41,25 +41,27 @@ class MoleculeConcurrentTest {
     var secondThread: Thread? = null
 
     val job = Job()
-    val cancelLatch = CompletableDeferred<Unit>()
-    launchMolecule(Immediate, job + Dispatchers.Default) {
-      var count by remember { mutableIntStateOf(0) }
-      when (count) {
-        0 -> firstThread = Thread.currentThread()
-        1 -> secondThread = Thread.currentThread()
+    try {
+      val cancelLatch = CompletableDeferred<Unit>()
+      launchMolecule(Immediate, job + Dispatchers.Default) {
+        var count by remember { mutableIntStateOf(0) }
+        when (count) {
+          0 -> firstThread = Thread.currentThread()
+          1 -> secondThread = Thread.currentThread()
+        }
+        if (count == 1) {
+          cancelLatch.complete(Unit)
+        }
+        SideEffect {
+          count++
+        }
       }
-      if (count == 1) {
-        cancelLatch.complete(Unit)
-      }
-      SideEffect {
-        count++
-      }
+      cancelLatch.await()
+
+      assertThat(firstThread).isSameInstanceAs(testThread)
+      assertThat(secondThread).isNotSameInstanceAs(testThread)
+    } finally {
+      job.cancelAndJoin()
     }
-    cancelLatch.await()
-
-    assertThat(firstThread).isSameInstanceAs(testThread)
-    assertThat(secondThread).isNotSameInstanceAs(testThread)
-
-    job.cancelAndJoin()
   }
 }
--- a/molecule-runtime/src/jvmTest/kotlin/app/cash/molecule/MoleculeConcurrentTest.kt
+++ b/molecule-runtime/src/jvmTest/kotlin/app/cash/molecule/MoleculeConcurrentTest.kt
@@ -29,19 +29,20 @@ import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.cancelAndJoin
-import kotlinx.coroutines.test.runTest
+import kotlinx.coroutines.runBlocking
 
 // These tests are JVM only because they look at the current thread ID. It could be supported on
 // all platforms with threads, but all the code is common, so this just gets us coverage quickly.
 class MoleculeConcurrentTest {
-  @Test fun coroutineContextHonoredByImmediateClock() = runTest {
+  @Test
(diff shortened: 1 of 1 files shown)


molecule-runtime/src/jvmTest/kotlin/app/cash/molecule/MoleculeConcurrentTest.kt
JakeWharton: I think this needs to be in a `try`/`finally` now.

With `runTest` the molecule coroutines were launched into the background scope and would be automatically canceled regardless of test outcome. We actually probably didn't need this line at all...

But with `runBlocking`, a failure will (probably) hang the suite since it will never exit. We need to ensure the job is torn down always. You should be able to test this by making one of the assertions fail temporarily and seeing what happens.

JakeWharton (approved):
Just one nit. Thanks!

molecule-runtime/src/jvmTest/kotlin/app/cash/molecule/MoleculeConcurrentTest.kt
FletchMcKee: I remember a while back looking into this and this pattern following the Recomposer tests had worked for me:

```kotlin
@Test fun coroutineContextHonoredByImmediateClock() = runTest(UnconfinedTestDispatcher()) {
  val testThread = Thread.currentThread()

  val job = Job()
  val threadChannel = Channel<Thread>(Channel.UNLIMITED)
  lateinit var recomposeScope: RecomposeScope
  backgroundScope.launchMolecule(Immediate, job + Dispatchers.Default) {
    threadChannel.trySend(Thread.currentThread())
    val scope = currentRecomposeScope
    SideEffect { recomposeScope = scope }
  }

  val firstThread = threadChannel.receive()
  assertThat(firstThread).isSameInstanceAs(testThread)

  recomposeScope.invalidate()

  val secondThread = threadChannel.receive()
  assertThat(secondThread).isNotSameInstanceAs(testThread)
  assertThat(secondThread.name).contains("DefaultDispatcher")

  job.cancelAndJoin()
}
```

This way you avoid reading `count` and writing to it in the same recomposer scope, but it’s been a while and it looks like they’ve changed those Recomposer tests.

FletchMcKee: Original [RecomposerTest](https://github.com/androidx/androidx/blob/cfeb69dddf7e86377e7052b3c8a968b31f6fcbf4/compose/runtime/runtime/src/nonEmulatorJvmTest/kotlin/androidx/compose/runtime/RecomposerTests.jvm.kt#L55-L90) for reference.

Amir-yazdanmanesh: > I think this needs to be in a `try`/`finally` now.
> 
> With `runTest` the molecule coroutines were launched into the background scope and would be automatically canceled regardless of test outcome. We actually probably didn't need this line at all...
> 
> But with `runBlocking`, a failure will (probably) hang the suite since it will never exit. We need to ensure the job is torn down always. You should be able to test this by making one of the assertions fail temporarily and seeing what happens.

  Good catch, wrapped the body in try/finally so job.cancelAndJoin() always runs, even if an assertion fails. This prevents the test from hanging the suite on failure.   
  I updated my PR

Amir-yazdanmanesh: > I remember a while back looking into this and this pattern following the Recomposer tests had worked for me:
> 
> ```kotlin
> @Test fun coroutineContextHonoredByImmediateClock() = runTest(UnconfinedTestDispatcher()) {
>   val testThread = Thread.currentThread()
> 
>   val job = Job()
>   val threadChannel = Channel<Thread>(Channel.UNLIMITED)
>   lateinit var recomposeScope: RecomposeScope
>   backgroundScope.launchMolecule(Immediate, job + Dispatchers.Default) {
>     threadChannel.trySend(Thread.currentThread())
>     val scope = currentRecomposeScope
>     SideEffect { recomposeScope = scope }
>   }
> 
>   val firstThread = threadChannel.receive()
>   assertThat(firstThread).isSameInstanceAs(testThread)
> 
>   recomposeScope.invalidate()
> 
>   val secondThread = threadChannel.receive()
>   assertThat(secondThread).isNotSameInstanceAs(testThread)
>   assertThat(secondThread.name).contains("DefaultDispatcher")
> 
>   job.cancelAndJoin()
> }
> ```
> 
> This way you avoid reading `count` and writing to it in the same recomposer scope, but it’s been a while and it looks like they’ve changed those Recomposer tests.

Thanks for the detailed diagnosis in the issue, it was really helpful in understanding the root cause. We went with runBlocking since the test genuinely needs real     
  Dispatchers.Default threads for thread identity checks, so there's no benefit to using a test dispatcher at all. The try/finally handles the cleanup concern Jake raised.

PR #733, opened 2026-02-15
 Add rememberDebounced composable utility     

### LINKED ISSUE #336: Debouncing state emissions
I'm trying to migrate
```
val data = snapshotFlow { buildFromState() }.debounce(1.seconds)
...
data.collectAsState(null)
```

to remove the view side default `null` value.

But there's no easy way to debounce in Molecule. I think the closest you can get would be
```
val debouncedFlow = snapshotFlow { buildFromState() }.debounce(1.seconds)
val state = scope.launchMolecule { 
    debouncedFlow.collectAsState(buildFromState()).value
}
```

but this moves from Compose (the actual state) -> Flow (`snapshotFlow`) -> Molecule -> Compose (`collectAsState`) -> Compose UI, which is very inefficient compared to reading the state directly inside Molecule. It also requires calling `buildFromState()` redundantly.

If there was a way to control/throttle the recomposition timing in Molecule, then I could try and make it only recompose every second or when the debounce concluded.

Amir-yazdanmanesh:
## Summary                                                                                                                                                               

  Closes #336

  Adds a `rememberDebounced<T>(value, timeout)` composable function to `molecule-runtime` that debounces state changes within Molecule compositions.

  - Returns the initial value immediately with no delay
  - On subsequent changes, delays propagation until the value has been stable for the full `timeout` duration
  - Resets the timer automatically if the value changes again before timeout expires (via `LaunchedEffect` cancellation)

  ## Test plan

  - [x] Initial value emitted immediately (ContextClock + Immediate)
  - [x] Value does not propagate before timeout elapses
  - [x] Value propagates after timeout elapses (ContextClock + Immediate)
  - [x] Rapid sequential changes — only last value propagates (ContextClock + Immediate)
  - [x] Zero-duration timeout behaves as immediate propagation
  - [x] All existing tests still pass
  - [x] API dump files updated and `apiCheck` passes


JakeWharton:
There's nothing specific to Molecule here. This is a general Compose problem that I don't think is our responsibility to solve.

Amir-yazdanmanesh:
> There's nothing specific to Molecule here. This is a general Compose problem that I don't think is our responsibility to solve.

  Fair point, thanks for the feedback. Closing this

PR #748, opened 2026-04-02
AGP 9 support

FletchMcKee:
Main changes are in molecule-runtime's build.gradle and the root build.gradle. Most info about the changes can be found [here](https://developer.android.com/kotlin/multiplatform/plugin). Basically moving the android block into kotlin block and replacing `androidTarget`. Also replacing `com.android.library` with `com.android.kotlin.multiplatform.library`. You can declare direct dependencies on `androidDeviceTest` now, so most of that logic got moved there while keeping the `kotlin-test-framework-junit` logic in the separate dependencies block.

Had to remove the BaseExtension logic from the root build.gradle. I tried to see if I could get something with CommonExtension to work, but ran into issues with the new `com.android.kotlin.multiplatform.library` plugin. So instead I separated this into two listeners for that plugin and `com.android.application`. Let me know if there is a better alternative.

The rest of the changes were removing the `org.jetbrains.kotlin.android` plugin from the samples while replacing variantFilters with androidComponents beforeVariants.

My groovy knowledge isn't the best anymore, so let me know if I'm using any incorrect syntax.

---

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,9 @@
+import com.android.build.api.dsl.ApplicationExtension
+import com.android.build.api.dsl.KotlinMultiplatformAndroidLibraryTarget
+
 buildscript {
   dependencies {
-    classpath libs.android.plugin
+    classpath libs.android.plugin.api
     classpath libs.kotlin.plugin.core
     classpath libs.kotlin.plugin.compose
     classpath libs.kotlin.plugin.serialization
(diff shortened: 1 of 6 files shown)


FletchMcKee:
I think we can also remove these from `gradle.properties` as well
```
android.defaults.buildfeatures.buildconfig=false
android.defaults.buildfeatures.aidl=false
android.defaults.buildfeatures.renderscript=false
```

build.gradle
```diff
@@ -47,23 +50,32 @@ subprojects {
...
-    def android = extensions.getByName("android") as com.android.build.gradle.BaseExtension
-    android.compileSdkVersion libs.versions.compileSdk.get().toInteger()
+  pluginManager.withPlugin('com.android.application') {
+    def android = extensions.getByName("android") as ApplicationExtension
+    android.compileSdk = libs.versions.compileSdk.get().toInteger()
     android.defaultConfig {
-      minSdkVersion libs.versions.minSdk.get().toInteger()
-      targetSdkVersion libs.versions.compileSdk.get().toInteger()
-    }
-    android.compileOptions {
```
FletchMcKee, line 57: I may need to add this back. Didn’t think it was necessary because of the `tasks.withType(JavaCompile).configureEach` block but it would probably be safer to set it here and in the `KotlinMultiplatformAndroidLibraryTarget` block as well.

PR #761, opened 2026-05-21
Fix launchMolecule startup cancellation race

### LINKED ISSUE #760: launchMolecule can call setContent on a disposed Composition during startup cancellation
### Description

`launchMolecule` can throw `IllegalStateException("The composition is disposed")` when the supplied coroutine context is cancelled during startup.

The current implementation creates a `Recomposer` and `Composition`, launches `recomposer.runRecomposeAndApplyChanges()` with `CoroutineStart.UNDISPATCHED`, disposes the composition in that coroutine's `finally`, and only then calls `composition.setContent { ... }`.

If the recomposer coroutine completes due to cancellation before `setContent` runs, `setContent` is invoked on an already-disposed composition.

I originally saw this in production Android crash reports with stacks like:

```text
java.lang.IllegalStateException: The composition is disposed
  at androidx.compose.runtime.CompositionImpl.ensureRunning(Composition.kt:1470)
  at androidx.compose.runtime.CompositionImpl.setContent(Composition.kt:634)
  at app.cash.molecule.MoleculeKt.launchMolecule(Molecule.kt:259)
  at app.cash.molecule.MoleculeKt.launchMolecule$default(Molecule.kt:217)
  at app.cash.molecule.MoleculeKt$immediateClockFlow$1$1$1.invokeSuspend(molecule.kt:85)
```

### Repro

This test demonstrates the failure on current `trunk` and also matches the same ordering in `2.2.0`:

```kotlin
@Test fun cancelledContextClockLaunch() = runTest {
  val job = Job()
  val clock = BroadcastFrameClock()
  val scope = CoroutineScope(coroutineContext + clock)

  job.cancel()

  assertFailure {
    scope.launchMolecule(ContextClock, emitter = { fail("emitter should not be called") }, context = job) {
      fail("body should not be composed")
    }
  }.hasMessage("The composition is disposed")
}
```

### Expected behavior

Launching with an already-cancelled context, or one that is cancelled during startup, should fail with normal coroutine cancellation semantics (`CancellationException`) and should not attempt `Composition.setContent` after disposal.

### Possible fix direction
A small fix might be to check `finalContext.ensureActive()` before creating/launching the composition, and/or reorder startup so `setContent` happens before the recomposer coroutine can dispose the composition. If reordering, startup failures before the recomposer coroutine is launched should still dispose the composition and snapshot observer before rethrowing.

jmalpasuto:
Fixes https://github.com/cashapp/molecule/issues/760

Ensures launchMolecule completes its synchronous initial composition and produces one value even when its launch context is already cancelled. Cancellation still prevents launched effects and future recompositions.

Keeps composition and snapshot observer cleanup tied to recomposer completion, and cancels the recomposer when initial composition or emitter code fails so startup resources do not leak.

### DIFF
--- a/molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
+++ b/molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
@@ -26,8 +26,6 @@ import kotlin.coroutines.CoroutineContext
 import kotlin.coroutines.EmptyCoroutineContext
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.CoroutineStart.UNDISPATCHED
-import kotlinx.coroutines.Job
-import kotlinx.coroutines.NonCancellable
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.coroutineScope
 import kotlinx.coroutines.flow.Flow
@@ -36,7 +34,6 @@ import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.channelFlow
 import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.launch
-import kotlinx.coroutines.withContext
 
 @Deprecated("", level = HIDDEN) // For binary compatibility.
 public fun <T> moleculeFlow(mode: RecompositionMode, body: @Composable () -> T): Flow<T> {
@@ -234,19 +231,8 @@ public fun <T> CoroutineScope.launchMolecule(
   val composition = Composition(UnitApplier, recomposer)
 
   var snapshotHandle: ObserverHandle? = null
-
-  val initialComposition = Job()
-  val runRecompose = launch(finalContext, start = UNDISPATCHED) {
-    try {
-      recomposer.runRecomposeAndApplyChanges()
-    } finally {
-      withContext(NonCancellable) {
-        initialComposition.join()
-        composition.dispose()
-        snapshotHandle?.dispose()
-        snapshotHandle = null
-      }
-    }
+  val recomposerJob = launch(finalContext, start = UNDISPATCHED) {
+    recomposer.runRecomposeAndApplyChanges()
   }
 
   try {
@@ -271,13 +257,13 @@ public fun <T> CoroutineScope.launchMolecule(
       emitter(body())
     }
   } catch (throwable: Throwable) {
-    snapshotHandle?.dispose()
-    snapshotHandle = null
     recomposer.cancel()
-    runRecompose.cancel()
     throw throwable
   } finally {
-    initialComposition.complete()
+    recomposerJob.invokeOnCompletion {
+      composition.dispose()
+      snapshotHandle?.dispose()
+    }
   }
 }
 
--- a/molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
+++ b/molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
@@ -223,11 +223,14 @@ public fun <T> CoroutineScope.launchMolecule(
   snapshotNotifier: SnapshotNotifier = defaultSnapshotNotifier(),
   body: @Composable () -> T,
 ) {
+  val baseContext = coroutineContext + context
+  if (!baseContext.isActive) return
+
   val clockContext = when (mode) {
     RecompositionMode.ContextClock -> EmptyCoroutineContext
     RecompositionMode.Immediate -> GatedFrameClock(this, context)
   }
-  val finalContext = coroutineContext + context + clockContext
+  val finalContext = baseContext + clockContext
 
   val recomposer = Recomposer(finalContext)
   val composition = Composition(UnitApplier, recomposer)
--- a/molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
+++ b/molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
@@ -24,16 +24,17 @@ import androidx.compose.runtime.snapshots.Snapshot
 import kotlin.DeprecationLevel.HIDDEN
 import kotlin.coroutines.CoroutineContext
 import kotlin.coroutines.EmptyCoroutineContext
+import kotlinx.coroutines.CancellationException
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.CoroutineStart.UNDISPATCHED
 import kotlinx.coroutines.channels.Channel
 import kotlinx.coroutines.coroutineScope
-import kotlinx.coroutines.ensureActive
 import kotlinx.coroutines.flow.Flow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.channelFlow
 import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.isActive
 import kotlinx.coroutines.launch
 
 @Deprecated("", level = HIDDEN) // For binary compatibility.
@@ -172,7 +173,7 @@ public fun <T> CoroutineScope.launchMolecule(
     body = body,
   )
 
-  return flow!!
+  return flow ?: throw CancellationException("launchMolecule was cancelled before producing an initial value")
 }
 
 @Deprecated("", level = HIDDEN) // For binary compatibility.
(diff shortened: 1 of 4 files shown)


molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
```diff
@@ -231,13 +232,8 @@ public fun <T> CoroutineScope.launchMolecule(
   val composition = Composition(UnitApplier, recomposer)
 
   var snapshotHandle: ObserverHandle? = null
-  launch(finalContext, start = UNDISPATCHED) {
-    try {
-      recomposer.runRecomposeAndApplyChanges()
-    } finally {
-      composition.dispose()
-      snapshotHandle?.dispose()
-    }
```
FletchMcKee, line 240: It's been a while since I looked through the Recomposer code, but I think this try/finally block needs to wrap `recomposer.runRecomposeAndApplyChanges()`. With this change if something throws in the `when (snapshotNotifier)` block, the snapshotHandle would leak and the composition wouldn't be disposed.

I think all you need to do is add the `finalContext.ensureActive()`.

jingibus: While I dislike silent failure as a failure mode, checking `isActive` and returning early would be idiomatic with how other launch methods behave on cancelled scopes. That is what I would do.

FletchMcKee: Yep, looking at some of the tests I wonder if there’s another issue. I remember discovering this snapshotHandle leak when working on #596 as I would see the errorDelayed tests handles leaking into other tests. That’s why I added this check to any of the tests that threw an exception:

```kotlin
// Verify `runRecomposeAndApplyChanges` is no longer active.
assertThat(job.isCompleted).isTrue()
```
But this didn’t work for the errorImmediately tests and I thought it wasn’t an issue because the exception was thrown in the initial composition so it never had a recomposition. For these tests I added `job.cancelAndJoin()` at the end and thought it was enough but I wonder if this is wrong. 

If an exception occurs in the initial composition, I don’t know if anything ever cleans up the snapshotHandle and runRecomposeAndApplyChanges ends up hanging. Maybe something like this is also needed?
```kotlin
if (!finalContext.isActive) return 
try {
  composition.setContent {
    emitter(body())
  }
} catch (exception: Exception) {
  finalContext.cancel()
  throw exception 
}
```
There’s gotta be a cleaner way and I may be overthinking this, maybe the job completes on the next frame clock tick, but some food for thought.

jmalpasuto: Thanks for reviewing! I put up a follow up fix that _hopefully_ addresses these issues... Let me know what you think!

molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
```diff
@@ -171,7 +173,7 @@ public fun <T> CoroutineScope.launchMolecule(
     body = body,
   )
 
-  return flow!!
+  return flow ?: throw CancellationException("launchMolecule was cancelled before producing an initial value")
```
jmalpasuto: I wasn't sure what to return if the flow is null now... CancellationException seemed to make sense?

JakeWharton: Hmm... What does `stateIn` do if you call it on a canceled scope? We should do whatever they do.

jmalpasuto: Good point... I'll take a peek

jmalpasuto: okay so looking at `stateIn`, the non-suspending [initial value overload](https://github.com/Kotlin/kotlinx.coroutines/blob/1.11.0/kotlinx-coroutines-core/common/src/flow/operators/Share.kt#L303-L312) can always return `MutableStateFlow(initialValue)`, even if the scope is cancelled.

the [suspending overload](https://github.com/Kotlin/kotlinx.coroutines/blob/1.11.0/kotlinx-coroutines-core/common/src/flow/operators/Share.kt#L322-L327) (with no initial value) waits for the first value via a `CompletableDeferred` parented to the scope job, so a cancelled scope throws `CancellationException`.

since Molecule has no initial value to return, I think the `StateFlow` overload should probably match the suspending `stateIn` behavior by throwing the `CancellationException`?

JakeWharton: The problem is that we aren't suspending, so we're not within a coroutine where that exception makes sense. Our function can be called synchronously on the main thread.

What if we wrap ourselves in `NonCancelable` and then cancel all of the children once we have our initial value?

JakeWharton: The assertion here is that we can always complete the initial recomposition to produce a single value, and then only then do we prevent the recomposer and effects from running and producing subsequent values.

jmalpasuto: good call. updated! if this isn't exactly what you're thinking lmk and / or feel free to make changes as you see fit

jmalpasuto: simplified this a bit, should be good for another pass

molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
```diff
@@ -231,34 +231,39 @@ public fun <T> CoroutineScope.launchMolecule(
...
     }
-  }
 
-  composition.setContent {
-    emitter(body())
+    composition.setContent {
+      emitter(body())
+    }
+  } catch (throwable: Throwable) {
+    recomposer.cancel()
```
jingibus, line 260: The `recomposer`, or the `recomposerJob`?

molecule-runtime/src/commonTest/kotlin/app/cash/molecule/MoleculeStateFlowTest.kt
```diff
@@ -99,10 +100,33 @@ class MoleculeStateFlowTest {
...
-    // This exception is processed in `composeInitial` and not `runRecomposeAndApplyChanges`, so the job is still active.
+    runCurrent()
+    assertThat(job.children.toList()).isEmpty()
     job.cancelAndJoin()
   }
 
+  @Test fun cancelledContextComposesInitialValueBeforeStopping() = runTest {
+    for (mode in listOf(ContextClock, Immediate)) {
+      val job = Job()
+      val scope = CoroutineScope(coroutineContext + BroadcastFrameClock())
```
jingibus, line 111: Breaks structured concurrency, which is bad — breaking SC can lead to unowned jobs, and concurrency leaks. 

Instead, do something like:

```kotlin
(this + BroadcastFrameClock()).launchMolecule<Int>(mode, context = job)
```

If that hangs, well — good! More coverage from the test.

molecule-runtime/src/commonTest/kotlin/app/cash/molecule/MoleculeTest.kt
```diff
@@ -110,10 +111,29 @@ class MoleculeTest {
...
-    // This exception is processed in `composeInitial` and not `runRecomposeAndApplyChanges`, so the job is still active.
+    runCurrent()
+    assertThat(job.children.toList()).isEmpty()
     job.cancelAndJoin()
   }
 
+  @Test fun cancelledContextComposesInitialValueBeforeStopping() = runTest {
+    for (mode in listOf(ContextClock, Immediate)) {
+      val job = Job()
+      val scope = CoroutineScope(coroutineContext + BroadcastFrameClock())
```
jingibus, line 122: Same feedback here.

molecule-runtime/src/commonTest/kotlin/app/cash/molecule/MoleculeStateFlowTest.kt
```diff
@@ -99,10 +100,33 @@ class MoleculeStateFlowTest {
...
+
+      val flow = scope.launchMolecule<Int>(mode, context = job) {
+        LaunchedEffect(Unit) {
+          effectRan = true
+        }
+        1
+      }
+      runCurrent()
+
+      assertThat(flow.value).isEqualTo(1)
```
jingibus, line 122: Move this before `runCurrent()` - this should synchronously be the case immediately after `launchMolecule`.

molecule-runtime/src/commonTest/kotlin/app/cash/molecule/MoleculeStateFlowTest.kt
jingibus: (This is contra the style used in other tests, I now realize, so - Jake will probably be fine with it)

jingibus (approved):

molecule-runtime/src/commonMain/kotlin/app/cash/molecule/molecule.kt
jmalpasuto: The recomposer. Canceling only the job could leave effects running whereas canceling the recomposer shuts down both.

molecule-runtime/src/commonTest/kotlin/app/cash/molecule/MoleculeStateFlowTest.kt
jmalpasuto: Good call checking this. The scope keeps runTest’s existing job, so no new unowned job is created.

molecule-runtime/src/commonTest/kotlin/app/cash/molecule/MoleculeStateFlowTest.kt
jmalpasuto: Updated!

molecule-runtime/src/commonTest/kotlin/app/cash/molecule/MoleculeStateFlowTest.kt
jingibus: Ahh, yeah, I missed that. I just automatically scan for CoroutineScope as being problematic. You are right!

jmalpasuto:
@jingibus @FletchMcKee @JakeWharton cool to merge?

FletchMcKee (approved):
Sorry meant to review this earlier. Looks like the snapshotHandle and composition are being disposed properly, not seeing it leak across tests so looks good to me.

sonrohan:
Any updates on this? Would be nice to have a fix.

jmalpasuto:
Awaiting a review from someone with write access

